PR #119, opened 2020-05-18
PL-58: Add AuditContext

effiban:
Adding the `AuditContext` interface which will be used to provide additional fields that should be added to the `AuditRecord` in order to tag/group the records (for example: profileId, advertiserId, etc.)
In this PR adding only the interface and corresponding annotation attribute.
The next PR-s will add the logic for populating the `AuditRecord` accordingly.


effiban:
@libbyf @mmarusich @amirz - FYI

PR #121, opened 2020-05-19
PL-60: Add ancestor fields to AuditedFieldSet

effiban:
Adding ancestor fields to `AuditedFieldSet`. These fields will be populated in the next PR by the resolver which will fetch them from the `AncestorFieldsProvider` (see also https://github.com/kenshoo/persistence-layer/pull/119 )


effiban:
@libbyf @mmarusich @amirz  - FYI

effiban:
retest this please

main/src/main/java/com/kenshoo/pl/entity/AncestorFieldsProvider.java
```diff
@@ -0,0 +1,28 @@
+package com.kenshoo.pl.entity;
```
galkoren, line 1: move to audit package?

effiban: I'm not sure about that - the current audit packages have internal code only, and this is part of the public API - maybe it should go under **spi** where I put the publisher? Or in a new package **spi.audit** ?

effiban: Same thing goes for `AuditRecord`

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
```diff
@@ -12,37 +12,75 @@
...
     private final Set<? extends EntityField<E, ?>> dataFields;
 
-    public AuditedFieldSet(final EntityField<E, ? extends Number> idField) {
-        this(idField, emptySet());
-    }
-    public AuditedFieldSet(final EntityField<E, ? extends Number> idField,
-                           final Iterable<? extends EntityField<E, ?>> dataFields) {
-        this.idField = requireNonNull(idField, "idField is required");
-        this.dataFields = dataFields == null ? emptySet() : ImmutableSet.copyOf(dataFields);
+    private AuditedFieldSet(final EntityField<E, ? extends Number> idField,
```
galkoren, line 24: Can we combine idField with ancestorFields into a single list ("keyFields)"?
Maybe the audit record shouldn't care if the key field comes from this entity or from an ancestor.

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
```diff
@@ -12,37 +12,75 @@
...
+    public static class Builder<E extends EntityType<E>> {
+        private final EntityField<E, ? extends Number> idField;
+        private Set<? extends EntityField<?, ?>> ancestorFields = emptySet();
+        private Set<? extends EntityField<E, ?>> dataFields = emptySet();
+
+        public Builder(EntityField<E, ? extends Number> idField) {
+            this.idField = requireNonNull(idField, "idField is required");
+        }
+
+        public Builder<E> withAncestorFields(final Iterable<? extends EntityField<?, ?>> ancestorFields) {
```
galkoren, line 71: This method is unused outside this class.
Shouldn't we use it in the resolver?

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
effiban: We can - but first of all I'm not sure if those fields are all going to have the meaning of "keys". Also I think there must be an id, it doesn't make sense that only the ancestor can identify you (then how are you a child?) . It gives a clear structure for both the Kafka message and the DB schema. If we mix in the id with the other keys it means that the clients will have to separate it from the other ones by themselves when they consume the data by the read API.

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
effiban: Yes, that will be in the next PR - trying to work with small PR-s now.

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
effiban: One more thing - we discussed that we may need to support multiple fields for the id (from the same entity) - so again I think it would be better to keep them separate from the ancestors.

main/src/main/java/com/kenshoo/pl/entity/AncestorFieldsProvider.java
galkoren: anywhere except for "entity" package, which is becoming bigger and bigger

effiban: Ok, I'll create an **audit** sub-package under every relevant package and move the code there

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
```diff
@@ -12,37 +12,75 @@
...
 
 import static java.util.Collections.emptySet;
+import static java.util.Collections.singleton;
 import static java.util.Objects.requireNonNull;
 
 public class AuditedFieldSet<E extends EntityType<E>> {
 
     private final EntityField<E, ? extends Number> idField;
+    private final Set<? extends EntityField<?, ?>> ancestorFields;
     private final Set<? extends EntityField<E, ?>> dataFields;
```
dimagorelik, line 22: @effiban 
Why the dates fields should have a special collections?

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
galkoren: 👍

PR #122, opened 2020-05-23
Allow Required(RELATION) in child command

galkoren:
## The Story

The ```@Required(RELATION)``` annotation has double meaning:
1. Validates that the field is populated on CREATE.
1. Allows the EntitiesFetcher to look up to ancestor fields.

So currently, when inserting a child (CampaignURL) to a parent (Campaign) like this :
```java
void test_insert_child() {
    pl.update(newCampaign().with(new CampaignURL().with(TEXT, "a.com"));
}
```

Using ```@Required(RELATION)``` will fail the validation, so this is not allowed:

```java
class CampaignURL {
    @Required(RELATION) // <------------- ah ah!
    public static EntityField<..., Integer> CAMPAIGN_ID = ...;
}
```

The validation fails because CAMPAIGN_ID will be populated only after the validation (by DbOutputGen), but the validator doesn't know it yet.

It is not a big issue. I can just remove the ```@Required``` annotation and assume no one will ever insert CampaignURL directly without a parent command.

But it is confusing. Developers came to me more than once with this issue.
Removing this limitation will also allow us to create a complete hierarchy (Campaign + AdGroup + Keyword) in a single command.
It will also allow me to write tests (Parent / Child / GrandChild) and define the relations correctly so that I can test GrandChild as standalone command that needs to fetch data from Parent.

## The fix
The fix is in the validations to ignore required fields when the command has a parent and the field is referencing the parent (and therefore shall be later populated).


main/src/main/java/com/kenshoo/pl/entity/internal/ChangesFilter.java
```diff
@@ -1,15 +1,12 @@
...
-import com.kenshoo.pl.entity.EntityType;
+import com.kenshoo.pl.entity.*;
 import com.kenshoo.pl.entity.spi.CurrentStateConsumer;
 
 import java.util.Collection;
 
 public interface ChangesFilter<E extends EntityType<E>> extends CurrentStateConsumer<E> {
 
-    <T extends EntityChange<E>> Collection<T> filter(Collection<T> changes, final ChangeOperation changeOperation, final ChangeContext changeContext);
+    <T extends ChangeEntityCommand<E>> Collection<T> filter(Collection<T> changes, final ChangeOperation changeOperation, final ChangeContext changeContext);
```
dimagorelik, line 10: @galkoren 
Why should we change the interface?  It should be immutable here

main/src/main/java/com/kenshoo/pl/entity/internal/IsFieldReferringToParentCommand.java
```diff
@@ -0,0 +1,34 @@
...
+
+        if (parentType == null) {
+            return whatever -> false;
+        }
+
+        return entityTypeOf(children).getKeyTo(parentType).from()::contains;
+    }
+
+    private static <E extends EntityType<E>> E entityTypeOf(Collection<? extends ChangeEntityCommand<E>> children) {
+        return seq(children).findFirst().get().getEntityType();
```
effiban, line 27: It's not good practice to call `get()` on an `Optional` without checking first (IntelliJ also warns about it).

Since you have two private methods which take the first child, you could add a method which only extracts the first child as an `Optional` and do it at the beginning of the chain in the `of()` method - then you won't need to keep fetching the first one from the collection.

main/src/main/java/com/kenshoo/pl/entity/internal/RequiredFieldsChangesFilter.java
```diff
@@ -25,14 +18,16 @@ public RequiredFieldsChangesFilter(final Set<EntityField<E, ?>> requiredFields)
     }
 
     @Override
-    public <T extends EntityChange<E>> Collection<T> filter(Collection<T> changes, ChangeOperation changeOperation, ChangeContext changeContext) {
-        return Collections2.filter(changes, change -> Iterables.all(requiredFields, entityField -> {
+    public <T extends ChangeEntityCommand<E>> Collection<T> filter(Collection<T> changes, ChangeOperation changeOperation, ChangeContext changeContext) {
+        Predicate<EntityField<E, ?>> isReferringToParentCommand = IsFieldReferringToParentCommand.of(changes);
+        return Collections2.filter(changes, change -> requiredFields.stream().allMatch(entityField -> {
```
effiban, line 23: This is not your code, but it would be nice to convert this to a Java 8 stream (the `Collections2.filter`) - maybe in a separate PR?

main/src/main/java/com/kenshoo/pl/entity/internal/RequiredFieldsCommandsFilter.java
```diff
@@ -22,13 +23,15 @@ public RequiredFieldsCommandsFilter(final Set<EntityField<E, ?>> requiredFields)
 
     @Override
     public <C extends ChangeEntityCommand<E>> Collection<C> filter(Collection<C> commands, final ChangeContext changeContext) {
+        Predicate<EntityField<E, ?>> isReferringToParentCommand = IsFieldReferringToParentCommand.of(commands);
         return Collections2.filter(commands, command -> Iterables.all(requiredFields, entityField -> {
             boolean fieldSpecified = command.isFieldChanged(entityField) && command.get(entityField) != null;
-            if (!fieldSpecified) {
+            boolean isValid = fieldSpecified || isReferringToParentCommand.test(entityField);
```
effiban, line 29: This OR condition now repeats twice - how about extracting to a shared method?

main/src/main/java/com/kenshoo/pl/entity/internal/RequiredFieldsCommandsFilter.java
```diff
@@ -22,13 +23,15 @@ public RequiredFieldsCommandsFilter(final Set<EntityField<E, ?>> requiredFields)
 
     @Override
     public <C extends ChangeEntityCommand<E>> Collection<C> filter(Collection<C> commands, final ChangeContext changeContext) {
+        Predicate<EntityField<E, ?>> isReferringToParentCommand = IsFieldReferringToParentCommand.of(commands);
         return Collections2.filter(commands, command -> Iterables.all(requiredFields, entityField -> {
```
dimagorelik, line 27: @galkoren 
Maybe we can use object Hierary from change context, what do you think?

main/src/main/java/com/kenshoo/pl/entity/internal/RequiredFieldsChangesFilter.java
galkoren: I see it accepts a lambda, so I'm not rush to replace it

effiban: Well it is still in the "Java 7" verbose style, which makes it harder to read, and less convenient to add additional operations - plus we are using Guava unnecessarily.
Not critical though.

main/src/main/java/com/kenshoo/pl/entity/EntityType.java
```diff
@@ -69,12 +69,12 @@ default SupportedChangeOperation getSupportedOperation() {
                 .findFirst();
     }
 
-    default Collection<EntityField<E, ?>> determineForeignKeys(Set<EntityField<E, ?>> requiredFields) {
+    default Stream<EntityField<E, ?>> determineForeignKeys(Set<EntityField<E, ?>> requiredFields) {
```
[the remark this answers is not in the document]
galkoren: I guarantee no one encountered this method in the KS.
If so, I will fix it in there.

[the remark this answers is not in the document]
galkoren: But now that we are open source, I should be more careful :)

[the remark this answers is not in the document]
effiban: :+1:

main/src/main/java/com/kenshoo/pl/entity/internal/IsFieldReferringToParentCommand.java
```diff
@@ -0,0 +1,30 @@
...
+                .orElse(anyField -> false);
+    }
+
+    private static <E extends EntityType<E>> Optional<EntityType.ForeignKey<E, ?>> keyToParent(ChangeEntityCommand<E> child) {
+        return ofNullable(child.getParent()).map(parent -> child.getEntityType().getKeyTo(parent.getEntityType()));
+    }
+
+    private static <T> Optional<T> first(Collection<T> items) {
+        return items.stream().findFirst();
+    }
```
effiban, line 28: :+1:

main/src/test/java/com/kenshoo/pl/entity/HierarchyTest.java
```diff
@@ -0,0 +1,81 @@
...
+        assertThat(hierarchy.getParent(Campaign).get(), is(Account));
+        assertThat(hierarchy.getParent(Sitelink).get(), is(Account));
+        assertThat(hierarchy.getParent(AdGroup).get(), is(Campaign));
+        assertThat(hierarchy.getParent(Ad).get(), is(AdGroup));
+        assertThat(hierarchy.getParent(SitelinkURL).get(), is(Sitelink));
+    }
+
+    @Test
+    public void root_parent_is_missing() {
+        assertFalse(hierarchy.getParent(Account).isPresent());
```
effiban, line 66: We have the **hamcrest-optional** library so you can use a more appropriate matcher:
```
import static com.github.npathai.hamcrestopt.OptionalMatchers.isPresent;
//...
assertThat(hierarchy.getParent(Account), isPresent());
```

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
```diff
@@ -226,7 +223,7 @@ private boolean isMissing(ChangeEntityCommand<?> cmd, ChangeContext context) {
 
     private <E extends EntityType<E>, C extends ChangeEntityCommand<E>> Collection<C> filterCommands(Collection<C> commands, ChangeFlowConfig<E> flowConfig, ChangeOperation changeOperation, ChangeContext changeContext) {
         if(changeOperation == CREATE) {
-            return Lists.newArrayList(new RequiredFieldsCommandsFilter<>(flowConfig.getRequiredRelationFields()).filter(commands, changeContext));
+            return Lists.newArrayList(new RequiredFieldsChangesFilter<>(flowConfig.getRequiredRelationFields()).filter(commands, changeOperation, changeContext));
```
dimagorelik, line 226: @galkoren 
Let's move this filter to flow config

PR #129, opened 2020-05-24
PL-62: Resolve 'always audited' fields and populate in AuditedFieldSet

effiban:
Continuation of https://github.com/kenshoo/persistence-layer/pull/121, but the naming has changed following today's meeting:
In `AuditFieldsResolver` add logic to resolve the "always audited" fields from the `@Audited` annotation class-level attribute, and populate them in the `AuditedFieldSet`

Next PR will complete the work, by enhancing the `AuditRecordGenerator` to populate these 'always' fields in the `AuditRecord`


effiban:
@libbyf @mmarusich @amirz - FYI

main/src/test/java/com/kenshoo/pl/entity/internal/audit/entitytypes/AuditedWithAncestorFieldsType.java
```diff
@@ -0,0 +1,37 @@
...
+import static com.kenshoo.pl.entity.annotation.RequiredFieldType.RELATION;
+
+@Audited(ancestorFieldsProvider = NotAuditedAncestorFieldsProvider.class)
+public class AuditedWithAncestorFieldsType extends AbstractEntityType<AuditedWithAncestorFieldsType> {
+
+    public static final AuditedWithAncestorFieldsType INSTANCE = new AuditedWithAncestorFieldsType();
+
+    @Id
+    public static final EntityField<AuditedWithAncestorFieldsType, Long> ID = INSTANCE.field(MainWithAncestorTable.INSTANCE.id);
+    @NotAudited
```
dimagorelik, line 22: @effiban 
Why this field shouldn't be audited?

main/src/test/java/com/kenshoo/pl/entity/internal/audit/entitytypes/AuditedWithAncestorFieldsType.java
```diff
@@ -0,0 +1,37 @@
...
+import com.kenshoo.pl.entity.annotation.Id;
+import com.kenshoo.pl.entity.annotation.Required;
+import com.kenshoo.pl.entity.annotation.audit.Audited;
+import com.kenshoo.pl.entity.annotation.audit.NotAudited;
+import com.kenshoo.pl.entity.internal.audit.MainWithAncestorTable;
+import com.kenshoo.pl.entity.internal.audit.ancestorfieldsproviders.NotAuditedAncestorFieldsProvider;
+
+import static com.kenshoo.pl.entity.annotation.RequiredFieldType.RELATION;
+
+@Audited(ancestorFieldsProvider = NotAuditedAncestorFieldsProvider.class)
```
dimagorelik, line 15: @effiban 
Does this annotation also provide additional fields from the entity itself or only a ancestor's fields?

main/src/test/java/com/kenshoo/pl/entity/internal/audit/entitytypes/AuditedWithAncestorFieldsType.java
effiban: First of all, in a real scenario it will not be necessary to audit this field since the ancestor id of an entity cannot be updated (usually) - plus we also have a separate mechanism for providing ancestor fields, which is the new parameter I just added.
In this case since it's only for testing, I wanted to keep it simple so I won't have to examine additional irrelevant fields in the assertion of the test.

main/src/test/java/com/kenshoo/pl/entity/internal/audit/entitytypes/AuditedWithAncestorFieldsType.java
effiban: It is supposed to be for ancestor fields only - but the interface is generic so theoretically someone could also add fields from the entity itself, although I don't see what they would gain from it.

dimagorelik: @effiban 
Maybe we should rename the interface?

effiban: What's wrong with the name? It's called `AnecstorFieldsProvider`

effiban: What I meant is that the **return value** of the interface is generic, not the name - we can't force someone to return fields from an ancestor as opposed to the entity itself (or else I don't know how to that)

effiban: I guess we could add a validation after the fields are fetched

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
```diff
@@ -18,40 +18,42 @@
...
+        this.onChangeFields = onChangeFields;
     }
 
     public EntityField<E, ? extends Number> getIdField() {
         return idField;
     }
 
-    public Set<? extends EntityField<?, ?>> getAncestorFields() {
-        return ancestorFields;
+    public Set<? extends EntityField<?, ?>> getAlwaysFields() {
```
dimagorelik, line 38: @effiban 
I did not find usage of this function getAlwaysFields(). It seems strange!

effiban: @dimagorelik - it will be in the next PR, when the generator will be changed to use these fields and populate them in the `AuditRecord`

PR #132, opened 2020-05-26
Fix entity change interface

dimagorelik:


main/src/main/java/com/kenshoo/pl/entity/internal/DbCommandsOutputGenerator.java
```diff
@@ -54,11 +54,12 @@ public void generate(Collection<? extends EntityChange<E>> entityChanges, Change
                                     changeContext,
                                     primaryTableCommands);
 
+                            //noinspection unchecked
                             new HierarchyKeyPopulator.Builder<E>()
                                     .with(changeContext.getHierarchy())
                                     .whereParentFieldsAre(autoInc())
                                     .gettingValues(fromContext(changeContext)).build()
-                                    .populateKeysToChildren(entityChanges);
+                                    .populateKeysToChildren((Collection<? extends ChangeEntityCommand<E>>)entityChanges);
```
dimagorelik, line 62: @galkoren @effiban 
We should rethink once more if it is a correct place to populate auto-generated ids.

PR #139, opened 2020-05-31
Pl context fetch by keys and pl condition API

dimagorelik:
Introduce new fetch by key API

Example:

final List<Entity> entities = plContext.select(TestEntityType.ID, TestEntityType.FIELD1)
                                               .from(TestEntityType.INSTANCE)
                                               .where(TestEntityType.FIELD1.eq("Alpha"))
                                               .fetchByKeys(keys);

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
@@ -137,7 +137,7 @@ public class PLContextSelectTest {
         final List<Entity> entities = plContext.select(TestEntityType.ID, TestEntityType.FIELD1)
                 .from(TestEntityType.INSTANCE)
                 .where(PLCondition.TrueCondition)
-                .fetch(ImmutableList.of(uniqueKey.createValue(1)));
+                .fetchByKeys(ImmutableList.of(uniqueKey.createValue(1)));
         assertThat("Incorrect number of entities fetched: ",
                 entities.size(), is(1));
 
--- a/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
@@ -1,5 +1,6 @@
 package com.kenshoo.pl.entity;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.kenshoo.jooq.AbstractDataTable;
 import com.kenshoo.jooq.DataTable;
@@ -128,6 +129,23 @@ public class PLContextSelectTest {
                                   fieldValue(TestParentEntityType.FIELD1, "ParentAlpha"),
                                   fieldValue(TestParentEntityType.SECONDARY_FIELD1, "ParentSecondaryAlpha")));
     }
+
+    @Test
+    public void selectFromSingleEntityByKeys() {
+        final SingleUniqueKey<TestEntityType, Integer> uniqueKey = new SingleUniqueKey<>(TestEntityType.ID);
+
+        final List<Entity> entities = plContext.select(TestEntityType.ID, TestEntityType.FIELD1)
+                .from(TestEntityType.INSTANCE)
+                .where(PLCondition.TrueCondition)
+                .fetch(ImmutableList.of(uniqueKey.createValue(1)));
+        assertThat("Incorrect number of entities fetched: ",
+                entities.size(), is(1));
+
+        assertThat(entities.get(0),
+                hasFieldValues(fieldValue(TestEntityType.ID, 1),
+                        fieldValue(TestEntityType.FIELD1, "Alpha")));
+    }
+
 
     @Test(expected = IllegalArgumentException.class)
     public void selectWithoutFieldsShouldThrowException() {
--- a/main/src/main/java/com/kenshoo/pl/entity/FetchFinalStep.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/FetchFinalStep.java
@@ -17,5 +17,5 @@ public interface FetchFinalStep {
      * @param keys the keys to fetch
      * @return the result entities
      */
-    List<Entity> fetch(Collection<? extends Identifier<?>> keys);
+    List<Entity> fetchByKeys(Collection<? extends Identifier<?>> keys);
 }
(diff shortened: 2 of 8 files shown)


main/src/main/java/com/kenshoo/pl/entity/FetchFinalStep.java
```diff
@@ -10,4 +11,11 @@
      * @return the result entities
      */
     List<Entity> fetch();
+
+    /**
+     * Finish building the query and fetch the results from the DB
+     *
+     * @return the result entities
+     */
+    List<Entity> fetch(Collection<? extends Identifier<?>> keys);
```
dimagorelik, line 20: @effiban @galkoren 
What do you think about the return value? 
Should it be a map or list?

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
```diff
@@ -36,14 +36,6 @@ public EntitiesFetcher(DSLContext dslContext, FeatureSet features) {
         this.oldEntityFetcher = new OldEntityFetcher(dslContext);
     }
 
-    public <E extends EntityType<E>> Map<Identifier<E>, Entity> fetchEntitiesByKeys(final E entityType,
```
dimagorelik, line 39: @galkoren 
I removed deprecated infterface

integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
```diff
@@ -129,6 +130,23 @@ public void selectFromChildAndParentAndParentSecondary() {
                                   fieldValue(TestParentEntityType.SECONDARY_FIELD1, "ParentSecondaryAlpha")));
     }
 
+    @Test
+    public void selectFromSingleEntityByKeys() {
+        final SingleUniqueKey<TestEntityType, Integer> uniqueKey = new SingleUniqueKey<>(TestEntityType.ID);
+
+        final List<Entity> entities = plContext.select(TestEntityType.ID, TestEntityType.FIELD1)
+                .from(TestEntityType.INSTANCE)
+                .where(PLCondition.TrueCondition)
```
dimagorelik, line 139: @effiban @galkoren 
Currently, the client code must define PL Condition. Should we do it optional?

main/src/main/java/com/kenshoo/pl/entity/FluentEntitiesFetcher.java
```diff
@@ -30,8 +35,10 @@ public FetchFinalStep where(final PLCondition condition) {
     }
 
     public List<Entity> fetch() {
-        return entitiesFetcher.fetch(entityType,
-                                     condition,
-                                     fieldsToFetch);
+        return entitiesFetcher.fetch(entityType, condition, fieldsToFetch);
+    }
+
+    public List<Entity> fetch(Collection<? extends Identifier<?>> keys) {
```
dimagorelik, line 41: @galkoren 
I think we should define generics here instead "?" because the entity type and the keys must be from the same type. What do you think?

main/src/main/java/com/kenshoo/pl/entity/FetchFinalStep.java
effiban, line 20: I think this approach of passing a **condition** on the keys to `fetch()` will be confusing to the user.
In regular Jooq when you pass fields to fetch it is like `select()` - indicating which fields to **return** so you can access them directly without going through a full record. 
However here it is used as a **filter** instead which belongs to the `where` clause (or else a `byKeys` clause which you started to write and we discussed before).
It also means we won't be able to use this method for filtering fields to be **returned**.

effiban: A list is fine. We can add an _additional_ method to fetch as a map, like Jooq does

integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
effiban: Yes - I think it should be optional

main/src/main/java/com/kenshoo/pl/entity/FetchFinalStep.java
galkoren: This is a condition and it should be added as a feature of PLCondition rather than a new separate fetching method.
Usage should be 
```java
PLContext
    .select(...)
    .from(...)
    .where(CampaignEntity.Key.in(campaignIds))
;
```
It might be tricky to support more than one key condition at the moment, so we can start with a limitation that throws an exception in such case.

galkoren: For performance,  we should return a Stream and let the user decide how to collect it.
We don't know in advance if the user wants a List or a Map.

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
```diff
@@ -71,6 +63,13 @@ public EntitiesFetcher(DSLContext dslContext, FeatureSet features) {
         return oldEntityFetcher.fetch(entityType, plCondition, fieldsToFetch);
     }
 
+    public List<Entity> fetch(final EntityType<?> entityType,
+                              final Collection<? extends Identifier<?>> ids,
+                              final PLCondition plCondition,
+                              final EntityField<?, ?>... fieldsToFetch) {
+        return oldEntityFetcher.fetch(entityType, ids, plCondition, fieldsToFetch);
```
galkoren, line 70: lets add new functionality to the new fetcher and not to the old one.

main/src/main/java/com/kenshoo/pl/entity/FetchFinalStep.java
dimagorelik: @galkoren 
The problem with CampaignEntity.Key.in(campaignIds) that we cannot support many of these conditions. It must be the only one.

galkoren: In the future, we could use QueryExtender within a loop. But the correct API can contain it right away.

integration-tests/src/test/java/com/kenshoo/pl/entity/PLContextSelectTest.java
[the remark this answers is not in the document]
effiban: Doesn't have to be in this PR, as discussed

PR #141, opened 2020-06-01
Add test for fetch by parent keys

dimagorelik:
Test for  fetch by parent keys

Example:

final List entities = plContext.select(TestEntityType.ID, TestEntityType.FIELD1)
.from(TestEntityType.INSTANCE)
.where(TestEntityType.FIELD1.eq("Alpha"))
.fetchByKeys(**parentKeys**)

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/fetch/OldEntityFetcher.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/fetch/OldEntityFetcher.java
@@ -17,6 +17,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.*;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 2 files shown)


main/src/main/java/com/kenshoo/pl/entity/internal/fetch/OldEntityFetcher.java
```diff
@@ -88,7 +89,8 @@ public OldEntityFetcher(DSLContext dslContext) {
         final AliasedKey<?> aliasedKey = new AliasedKey<>(uniqueKey);
 
         final Set<EntityField<?, ?>> requestedFieldsToFetch = ImmutableSet.copyOf(fieldsToFetch);
-        final Set<? extends EntityField<?, ?>> allFieldsToFetch = Sets.union(requestedFieldsToFetch, plCondition.getFields());
```
dimagorelik, line 91: @effiban @galkoren 
Do you know the better way to contact 3 collections into one set?

effiban: @dimagorelik - you can do it like this:
```
Stream.of(stream1, stream2, stream3)
      .flatMap(Function.identity())
      .collect(toSet());
```
                 
```

dimagorelik: @effiban 
It does not work because the order is important.

PR #143, opened 2020-06-02
PL-69: Modify generator to populate mandatory fields in AuditRecord

effiban:
This is the continuation of https://github.com/kenshoo/persistence-layer/pull/129 and https://github.com/kenshoo/persistence-layer/pull/144.
Modifying the `AuditRecordGenerator` to take the mandatory fields from the `AuditedFieldSet` and populate them in the `AuditRecord`.

### DIFF
--- a/integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateOneLevelWithMandatoryTest.java
+++ b/integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateOneLevelWithMandatoryTest.java
@@ -17,9 +17,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.util.List;
-import java.util.Objects;
 
-import static com.kenshoo.pl.entity.ChangeOperation.CREATE;
 import static com.kenshoo.pl.entity.matchers.audit.AuditRecordMatchers.*;
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
@@ -35,15 +33,22 @@ public class AuditForCreateOneLevelWithMandatoryTest {
     private static final String ANCESTOR_NAME = "ancestorName";
     private static final String ANCESTOR_DESC = "ancestorDesc";
 
+    private static final EntityField<AuditedWithAncestorMandatoryType, Long> ANCESTOR_FK_FIELD = AuditedWithAncestorMandatoryType.ANCESTOR_ID;
+    private static final EntityField<AuditedWithAncestorMandatoryType, String> NAME_FIELD = AuditedWithAncestorMandatoryType.NAME;
+    private static final EntityField<AuditedWithAncestorMandatoryType, String> DESC_FIELD = AuditedWithAncestorMandatoryType.DESC;
+
+    private static final EntityField<NotAuditedAncestorType, String> ANCESTOR_NAME_FIELD = NotAuditedAncestorType.NAME;
+    private static final EntityField<NotAuditedAncestorType, String> ANCESTOR_DESC_FIELD = NotAuditedAncestorType.DESC;
+
     private static final List<DataTable> ALL_TABLES = ImmutableList.of(MainWithAncestorTable.INSTANCE,
                                                                        AncestorTable.INSTANCE);
 
     private PLContext plContext;
     private InMemoryAuditRecordPublisher auditRecordPublisher;
 
-    private ChangeFlowConfig<AuditedWithAncestorMandatoryType> auditedWithAncestorMandatoryConfig;
+    private ChangeFlowConfig<AuditedWithAncestorMandatoryType> mainWithAncestorMandatoryConfig;
 
-    private PersistenceLayer<AuditedWithAncestorMandatoryType> auditedWithAncestorMandatoryPL;
+    private PersistenceLayer<AuditedWithAncestorMandatoryType> mainWithAncestorMandatoryPL;
 
     @Before
     public void setUp() {
@@ -54,9 +59,9 @@ public class AuditForCreateOneLevelWithMandatoryTest {
             .withAuditRecordPublisher(auditRecordPublisher)
             .build();
 
-        auditedWithAncestorMandatoryConfig = flowConfig(AuditedWithAncestorMandatoryType.INSTANCE);
+        mainWithAncestorMandatoryConfig = flowConfig(AuditedWithAncestorMandatoryType.INSTANCE);
 
-        auditedWithAncestorMandatoryPL = persistenceLayer();
+        mainWithAncestorMandatoryPL = persistenceLayer();
 
         ALL_TABLES.forEach(table -> DataTableUtils.createTable(dslContext, table));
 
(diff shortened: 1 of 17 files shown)


effiban:
@amirz @mmarusich @libbyf @UncleTalik - FYI

integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateOneLevelWithMandatoryTest.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        ALL_TABLES.forEach(table -> plContext.dslContext().dropTable(table).execute());
+    }
+
+    @Test
+    public void oneEntity_WithEntityLevelMandatoryFields_AndFieldsInCmd_ShouldCreateMandatoryFieldsAndFieldRecords() {
+        final CreateResult<AuditedWithAncestorMandatoryType, Identifier<AuditedWithAncestorMandatoryType>> createResult =
```
galkoren, line 77: Welcome to my **Shorting-The-Lines** show!
(You don't have to agree. I will respect other opinion).
* We can ```import static AuditedWithAncestorMandatoryType.*``` to we can remove this long prefix from the field names.
* We can extract the expression ```blablaPL.create(.., flowConfig)``` into a method that returns the ID.
* Can extract the long expression ```new CreateAuditedWithAncestorMandatoryCommand()``` into a method with a short name.

Tell me whether this is more readable:
```java
final long id = execute(createCommand()
    .with(ANCESTOR_ID, ANCESTOR_ID_VALUE)
    .with(NAME, NAME_VALUE)
    .with(DESC, DESC_VALUE)
);

assertThat(capturedAudits(), hasSize(1));

assertThat(capturedAudits().get(0), allOf(hasEntityType(ENTITY_TYPE),
                                      hasEntityId(String.valueOf(id)),
                                      // the following 2 lines cannot get shortened... 
                                      // unless.. you rename these fields to be unique so I can static import them too
                                      hasMandatoryFieldValue(NotAuditedAncestorType.NAME, ANCESTOR_NAME),
                                      hasMandatoryFieldValue(NotAuditedAncestorType.DESC, ANCESTOR_DESC),
                                      hasOperator(CREATE),
                                      hasCreatedFieldRecord(NAME, NAME_VALUE),
                                      hasCreatedFieldRecord(DESC, DESC_VALUE)));




```

integration-tests/src/test/java/com/kenshoo/pl/audit/commands/CreateAuditedWithAncestorMandatoryCommand.java
```diff
@@ -0,0 +1,13 @@
+package com.kenshoo.pl.audit.commands;
+
+import com.kenshoo.pl.entity.CreateEntityCommand;
+import com.kenshoo.pl.entity.EntityCommandExt;
+import com.kenshoo.pl.entity.internal.audit.entitytypes.AuditedWithAncestorMandatoryType;
+
+public class CreateAuditedWithAncestorMandatoryCommand extends CreateEntityCommand<AuditedWithAncestorMandatoryType>
+    implements EntityCommandExt<AuditedWithAncestorMandatoryType, CreateAuditedWithAncestorMandatoryCommand> {
```
galkoren, line 8: This ```EntityCommandExt``` of yours changed my life 👍

main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
```diff
@@ -1,31 +1,38 @@
...
 
 import static java.util.Collections.emptyList;
+import static java.util.Collections.emptyMap;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 
 public class AuditRecord<E extends EntityType<E>> {
     private final E entityType;
     private final String entityId;
+    private final Map<? extends EntityField<?, ?>, ?> mandatoryFieldValues;
```
galkoren, line 22: Map has a large memory overhead ..
Let's make it a ```Collection<Pair<EntityField, ?>>```.
Or we can define a class and make it ```Collection<FieldAndValue>```.
(actually, we already have a class FieldAndValue in the "jooq" package, which is not good for us)

galkoren: we can also shorten the name to "mandatoryValues"

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
```diff
@@ -61,6 +66,8 @@ public AuditRecordGenerator(final AuditedFieldSet<E> auditedFieldSet) {
 
         final String entityId = extractEntityId(entityChange, entity);
 
+        final Map<? extends EntityField<?, ?>, ?> mandatoryFieldValues = generateMandatoryFieldValues(entity);
```
galkoren, line 69: save as [here](https://github.com/kenshoo/persistence-layer/pull/143/files#r434848922)

main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
galkoren: Actually, when Victor was here, he preferred arrays over collections :)

integration-tests/src/test/java/com/kenshoo/pl/audit/AuditForCreateOneLevelWithMandatoryTest.java
effiban: I agree with most of what you say - indeed the lines are long, but I have two comments:
1) Regarding static imports: I don't think that's a good idea to do exactly as you said,  because this would create a "coupling" between two classes that are forced to coordinate their names so they don't clash - and these classes are being used by multiple test classes. 
It would be nice if we had a `typedef` like C or Scala to solve this, but since we don't I can instead create a "fake typedef" by creating an empty inner class extending the original one:
```
private static class AncestorType extends NotAuditedAncestorType {
}
```
- and then I can use it instead to shorten the name, or we could even add some aliases for the fields inside to change their names and then import _them_ statically.

2) Regarding the extra assertions - I thought about that for a while. In some of the tests (also from previous PR-s) you'll see that not all the fields are being asserted, however I also think there is value in checking fields which are supposedly irrelevant, because I want to treat the code as a black box and make sure that it doesn't break in the future. There could be a bug where fields which are unrelated (like non-mandatory fields) will be excluded because the mandatory is there and some condition in the code got it wrong.
Specifically here I would keep the assertions on the **field records** since they are the main part of the auditing, and remove the entity type/id/operator which do indeed repeat often in other tests.

integration-tests/src/test/java/com/kenshoo/pl/audit/commands/CreateAuditedWithAncestorMandatoryCommand.java
effiban: I think we all did it together - but thanks :)

main/src/main/java/com/kenshoo/pl/entity/audit/AuditRecord.java
effiban: OK you're right that the collection (List) will take less memory, even though we are talking about small `Map`s - but I guess it adds up with all the entities in a given flow.
Yes there are a few FieldAndValue classes but the one we need is only defined for tests, I will change that.
Regarding arrays.. :fearful: -  an ArrayList (for instance) starts out with a capacity of 10 and the difference from an array should be negligible. Or we can create it with the exact capacity.  I also just saw there's a method trimToSize() which remove extra allocated space. So I think Java collections are well-built and we don't need to worry too much.
I admit I'm not good at analyzing performance, but I think sometimes we get carried away... let's add performance tests to the repo so we can check these things properly (I'd like to do that next when I have time...)

effiban: Regarding the name - I prefer `mandatoryFieldValues` because it should be clear that these are pairs and not just the values (I would find it confusing to shorten this).

effiban:
@galkoren - fixed your comments

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
effiban: Answered above - I prefer the current name

PR #144, opened 2020-06-02
PL-72: Rename 'alwaysAudited' to 'mandatory'

effiban:
The `alwaysAudited`/`always` naming didn't quite fit when I carried it on to the next task of modifying `AuditRecordGenerator` and `AuditRecord`.
It's not quite right in English and is confusing + causes long names.

So this PR renames it to just 'mandatory'. The word 'audited' is removed where it can be inferred from the context


effiban:
@mmarusich @libbyf @amirz - FYI

PR #146, opened 2020-06-04
Add isNull condition

yoelg:
Add isNull condition


main/src/main/java/com/kenshoo/pl/entity/EntityField.java
```diff
@@ -43,4 +43,13 @@ default PLCondition in(T ...values) {
         final TableField<Record, Object> tableField = (TableField<Record, Object>)getDbAdapter().getFirstTableField();
         return new PLCondition(tableField.in(tableValues), this);
     }
+
+    default PLCondition isNull() {
+        if (isVirtual()) {
+            throw new UnsupportedOperationException("The equals operation is unsupported for virtual fields");
```
dimagorelik, line 49: @galkoren 
The isNull operation is unsupported for virtual fields

PR #147, opened 2020-06-04
PL-73: Refactor MandatoryFieldsProvider to AuditExtensions

effiban:
Following my conversation with @galkoren -  refactoring `MandatoryFieldsProvider` to `AuditExtensions`.
Apart from the name change, this gives us a more flexible interface that we can extend in the future. The mandatory fields from external entities are now only one kind of extension, and we can add more in the future to provide additional extensions to the data that may be required.
In addition it makes the naming conventions simpler and cleaner going forward.


effiban:
@mmarusich @libbyf @amirz - FYI

PR #148, opened 2020-06-07
PL-70: Auditing: Support mandatory fields from the same entity -  part 1: Resolve annotations

effiban:
Adding support for mandatory audit fields coming from the same entity (in addition to **external** mandatory fields already supported).
This addresses a Kenshoo requirement for including the "match type" of a keyword in all the audit records, even if it hasn't changed.

Implemented as a new attribute 'trigger' of the `@Audited` annotation which can either be 'ALWAYS' for mandatory fields or 'ON_CHANGE' for others (default). 
Example:
```
@Audited
public class KeywordType extends AbstractEntityType<KeywordType> {

    @Audited(trigger = ALWAYS)
    public static final EntityField<KeywordType, String> MATCH_TYPE = INSTANCE.field(KeywordTable.INSTANCE.match_type);
}
```

**This is part 1 of the feature** which only resolves the annotations and populates them in the `AuditedFieldSet` (for each flowConfig).
Since the resolution logic got more complicated, I separate the existing resolver into `SelfAuditedFieldsResolver` which handles the fields of the same entity, and `AllAuditedFieldsResolver` which calls the former and adds the external mandatory fields as well.

Part 2 will modify the `AuditRecordGenerator` to populate these fields in the `AuditRecord` (for each entity in each command).


effiban:
@mmarusich @amirz @libbyf - FYI

main/src/main/java/com/kenshoo/pl/entity/internal/audit/SelfAuditedFieldsResolver.java
```diff
@@ -0,0 +1,98 @@
...
+
+        final Optional<AuditTrigger> optionalEntityTrigger = extractEntityTrigger(entityType);
+        final Collection<AuditedFieldTrigger<E>> fieldTriggers = resolveFieldTriggers(nonIdFields, () -> optionalEntityTrigger);
+
+        final AuditedFieldSet<E> fieldSet = AuditedFieldSet.builder(idField)
+                                                           .withSelfMandatoryFields(filterFieldsByTrigger(fieldTriggers, ALWAYS))
+                                                           .withOnChangeFields(filterFieldsByTrigger(fieldTriggers, ON_CHANGE))
+                                                           .build();
+
+        if (shouldAudit(entityType, fieldSet)) {
```
galkoren, line 45: 1. bug
1. readability:
```java
return isAudited(entity) || fieldSet.notEmpty()
    ? Optional.of(result)
    : Optional.empty();
```

main/src/main/java/com/kenshoo/pl/entity/internal/audit/SelfAuditedFieldsResolver.java
```diff
@@ -0,0 +1,98 @@
...
+        requireNonNull(entityType, "entityType is required");
+        return entityType.getIdField()
+                         .flatMap(idField -> resolve(entityType, idField));
+    }
+
+    private <E extends EntityType<E>> Optional<AuditedFieldSet<E>> resolve(final E entityType, final EntityField<E, ? extends Number> idField) {
+        final Stream<? extends EntityField<E, ?>> nonIdFields = entityType.getFields()
+                                                                          .filter(field -> !idField.equals(field));
+
+        final Optional<AuditTrigger> optionalEntityTrigger = extractEntityTrigger(entityType);
```
galkoren, line 37: return to boolean

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldTrigger.java
```diff
@@ -0,0 +1,23 @@
+package com.kenshoo.pl.entity.internal.audit;
+
+import com.kenshoo.pl.entity.EntityField;
+import com.kenshoo.pl.entity.EntityType;
+import com.kenshoo.pl.entity.audit.AuditTrigger;
+
+class AuditedFieldTrigger<E extends EntityType<E>> {
```
galkoren, line 7: shorter name: AuditedField (and "trigger" can be a property)

effiban: :+1:

effiban: actually not sure about this - because we have `AuditedFieldSet` as well and they don't exactly correspond to each other

effiban:
@galkoren - comments fixed

PR #149, opened 2020-06-13
PL-75: Simplify Audit fields resolver

effiban:
Following comments from the previous PR:

- Replacing the 2 audit fields resolvers with a single **AuditedFieldsResolver**  with no interface (**this will be hard to see in the diffs** because it looks like an interface became a class. Compare it to the old "SelfAuditedFieldsResolver")
- **AuditedFieldsResolver** will call a new class **ExternalMandatoryFieldsExtractor** to get the external fields (instead of calling an inner resolver to get the "self" fields)


effiban:
@amirz - FYI

effiban:
@galkoren - please review again, I did it in a simpler way

main/src/main/java/com/kenshoo/pl/entity/internal/audit/ExternalMandatoryFieldsExtractor.java
```diff
@@ -12,33 +11,13 @@
 import java.util.Optional;
 import java.util.stream.Stream;
 
-import static org.jooq.lambda.Seq.seq;
+public class ExternalMandatoryFieldsExtractor {
```
galkoren, line 14: 👍

PR #150, opened 2020-06-21
Add Entity.getOptional() to replace multiple private methods

effiban:
This refactoring reduces code duplication and is also a preparation for my next PR related to auditing


effiban:
@amirz - FYI

PR #151, opened 2020-06-21
Add FieldsValueMap.getOptional() to replace multiple private methods

effiban:
Similar to https://github.com/kenshoo/persistence-layer/pull/150 for `FieldsValueMap` (parent of `EntityChange` and many others)


main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
```diff
@@ -95,7 +95,7 @@ public AuditRecordGenerator(final AuditedFieldSet<E> auditedFieldSet) {
                                                        final EntityField<E, ?> field) {
         return new FieldAuditRecord<>(field,
                                       entity.getOptional(field).orElse(null),
-                                      entityChange.get(field));
+                                      entityChange.getOptional(field).orElse(null));
```
effiban, line 98: I will add a builder for `FieldAuditRecord` later on, and then both of the `orElse`-s above will not be needed.

effiban:
**General comment:** 
Since `Entity` and `FieldsValueMap` both allow `null` values, we won't be able to get rid of the `containsField` methods right away, even if we do get rid of `get`.
I think we can consider a better (and 'functional'-friendly) of indicating that a field is empty, without saving `null` values in a `Map` - maybe by creating a separate collection of "empty" or "cleared" fields instead of storing them all in the same `Map` - or else defining an empty value type for each type.
Obviously this will not be an easy refactoring and we will probably not be able to do it anytime soon...

effiban:
@amirz  - FYI

main/src/test/java/com/kenshoo/pl/entity/FieldsValueMapTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private static final String DUMMY_FIELD_VALUE = "abc";
+
+    @SuppressWarnings("unchecked")
+    @Mock
+    final EntityField<TestEntity, String> mockField = mock(EntityField.class);
+
+    @Test
+    public void testGetOptional_WhenNotNull_ShouldReturnIt() {
+        final FieldsValueMap<TestEntity> fieldsValueMap = new FieldsValueMap<TestEntity>() {
```
galkoren, line 21: Instead of defining an anonymous class in each test, for the same amount of code, you can define a real inner class:
```java
class TestableMap implements FieldsValueMap {
    Map<EntityField, Object> map;
    FieldsValueMap (Map<EntityField, Object> map) { this.map = map; }
}
```

So the test scenarios could look like this:
```java
var map = new TestableMap(ImmutableMap.of(DUMMY_FIELD, 55555));
assertThat(map.getOptional(DUMMY_FIELD), isPresentAndIs(55555));
```

effiban: :+1:

effiban:
Will redo this with Triptional

PR #152, opened 2020-06-22
Add Triptional (three-state object)

effiban:
In the PL we have a notion of a three-state object that can be one of the following:
- Present with non-`null` value
- Present with `null` value
- Absent

This is used when considering a field of an entity in 2 cases:
- in `Entity` after being fetched from the DB, it can be either:
  - Fetched with a non-`null` value
  - Fetched with a `null` value
  - Not fetched

- In `EntityChange` (command) it can be either:
  - Changed in the command to a non-`null` value
  - Changed in the command to a `null` value
  - Not in the command (unchanged)

Currently we represent this by a `containsField()` method together with a `get()` method that can return `null` and that's not quite the approach for functional programming, and in general is not easy to use. 
We could handle this with additional methods or a wrapper "nullable" object for the value but that can be cumbersome. So instead I am adding a three-state object that will represent these cases.

Credit to @galkoren for the name :)

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/TriState.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/TriState.java
@@ -1,0 +1,107 @@
+package com.kenshoo.pl.entity;
+
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static com.kenshoo.pl.entity.TriState.State.*;
+import static java.util.Objects.requireNonNull;
+import static java.util.function.Function.identity;
+
+public class TriState<T> {
+
+    enum State {
+        FILLED,
+        EMPTY,
+        ABSENT
+    }
+
+    private static final TriState<?> EMPTY_INSTANCE = new TriState<>(EMPTY);
+    private static final TriState<?> ABSENT_INSTANCE = new TriState<>(ABSENT);
+
+    private final T value;
+    private final State state;
+
+    private TriState(final State state) {
+        this(null, state);
+    }
+
+    private TriState(final T value) {
+        this(value,
+             value == null ? EMPTY : FILLED);
+    }
+
+    private TriState(final T value, final State state) {
+        this.value = value;
+        this.state = state;
+    }
+
+    public static <T> TriState<T> of(final T value) {
+        return new TriState<>(value);
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T> TriState<T> empty() {
+        return (TriState<T>)EMPTY_INSTANCE;
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T> TriState<T> absent() {
+        return (TriState<T>)ABSENT_INSTANCE;
+    }
+
+    public <U> TriState<U> map(final Function<? super T, ? extends U> mapper) {
+        return map(mapper, () -> null);
+    }
+
+    public <U> TriState<U> map(final Function<? super T, ? extends U> filledMapper,
+                               final Supplier<? extends U> emptySupplier) {
+        requireNonNull(filledMapper, "filledMapper is required");
+        requireNonNull(emptySupplier, "emptySupplier is required");
+
+        switch (state) {
+            case FILLED:
+                return of(filledMapper.apply(value));
+            case EMPTY:
+                return of(emptySupplier.get());
+            default:
+                return absent();
+        }
+    }
+
+    public Optional<T> asOptional() {
+        return mapToOptional(identity());
+    }
+
+    public <U> Optional<U> mapToOptional(final Function<? super T, ? extends U> mapper) {
+        return mapToOptional(mapper, () -> null);
+    }
+
+    public <U> Optional<U> mapToOptional(final Function<? super T, ? extends U> filledMapper,
+                                         final Supplier<? extends U> emptySupplier) {
+        switch (state) {
+            case FILLED:
+                return Optional.ofNullable(filledMapper.apply(value));
+            case EMPTY:
+                return Optional.ofNullable(emptySupplier.get());
+            default:
+                return Optional.empty();
+        }
+    }
+
+    public boolean isFilled() {
+        return state == FILLED;
+    }
+
+    public boolean isNotFilled() {
+        return !isFilled();
+    }
+
+    public boolean isEmpty() {
+        return state == EMPTY;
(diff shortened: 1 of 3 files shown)


main/src/main/java/com/kenshoo/pl/entity/TriState.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static com.kenshoo.pl.entity.TriState.State.*;
+import static java.util.Objects.requireNonNull;
+import static java.util.function.Function.identity;
+
+public class TriState<T> {
```
dimagorelik, line 11: @effiban 
I am not sure that the usage of this class will be simpler than contains field and get API

effiban: It will allow you to do it in functional style and not require a separate function to encapsulate the "if":

// CURRENT approach
```
fields.stream()
        .map(f -> getEntityValue(f))
        .filter(Optional::isPresent)
        .map(Optional::get)
        .forEach(f -> writeAuditRow(f)));
//...
private Optional<?> getEntityValue(field) {
    if (entity.containsField) {
         return Optional.ofNullable(entity.get(field)); 
    }
    return Optional.empty();
}
```

// NEW approach
```
fields.stream()
        .map(f -> entity.getTriValue(f))
        .filter(TriState::isFilled)
        .map(TriState::get)
        .forEach(f -> writeAuditRow(f)));
```

effiban: At the same time, in another part of the code you can still check if the field exists (state != ABSENT, I need to add that method), and do something if it is "empty" (meaning exists and is `null`)

effiban: We could even create shortcut methods for this in `Entity` or `EntityChange`. 
It will be short and it won't break any existing code.

main/src/main/java/com/kenshoo/pl/entity/Triptional.java
```diff
@@ -0,0 +1,111 @@
...
+        this.value = value;
+        this.state = state;
+    }
+
+    public static <T> Triptional<T> of(final T value) {
+        return new Triptional<>(value);
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T> Triptional<T> nul() {
```
effiban, line 44: With one 'l' since `null` is an invalid name

galkoren: prefer NULL (capital) or null_instance or the_null...

main/src/main/java/com/kenshoo/pl/entity/Triptional.java
```diff
@@ -0,0 +1,111 @@
...
+                return absent();
+        }
+    }
+
+    public Optional<T> asOptional() {
+        return mapToOptional(identity());
+    }
+
+    public <U> Optional<U> mapToOptional(final Function<? super T, ? extends U> mapper) {
+        return mapToOptional(mapper, () -> null);
```
galkoren, line 77: 1. This will throw an exception if `null`.
1. Unit test is required

main/src/main/java/com/kenshoo/pl/entity/Triptional.java
effiban: None of these follow the Java standard :) but I can do `nullInstance`. I was trying to keep it short...

main/src/main/java/com/kenshoo/pl/entity/Triptional.java
effiban: There are no unit tests yet, I wanted to make sure we agree on this class in general

main/src/main/java/com/kenshoo/pl/entity/Triptional.java
```diff
@@ -0,0 +1,151 @@
...
+                return of(filledMapper.apply(value));
+            case NULL:
+                return of(nullReplacer.get());
+            default:
+                return absent();
+        }
+    }
+
+    public Optional<T> asOptional() {
+        return Optional.ofNullable(value);
```
galkoren, line 83: We have:
* asOptional
* mapToOptional

One of them supposed to consider null as empty and the other need not?

effiban: The default (`asOptional`) is to consider null as empty and not change anything else.
The 2 `mapToOptional`-s allow you different behavior: You can replace the value if not-null (by the mapper), **and/or** you can replace the `null` with a not-null value.

I'm not sure exactly which of the methods will be useful (and maybe we can remove some later) - but I tried to create something flexible enough.

effiban:
@amirz - FYI

main/src/main/java/com/kenshoo/pl/entity/Triptional.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static com.kenshoo.pl.entity.Triptional.State.*;
+import static java.util.Objects.requireNonNull;
+
+public class Triptional<T> {
+
+    enum State {
+        FILLED,
```
amirz, line 17: Would go for 'HAS_VALUE' or 'NOT_NULL'

effiban: Either one is fine as an enum value, however I want to be clear on the distinction between the three states.
There's a corresponding accessor `isFilled()` that will have to be renamed and then it might not be clear:

- If we rename to `hasValue()` it might not be clear that this doesn't include `null`.
- If we rename to `isNotNull()` - then technically, it could also mean "absent" even though that's not what we want.
There's also a negative accessor `isNotFilled()` and it won't be clear how to rename that.

@galkoren - WDYT?

galkoren: I vote for "NOT_NULL"

effiban: Ok, I need to think some more about how this can fit in with the relevant methods.
There's also Javadoc I have to add anyway,
In the meantime will merge so I can continue with changes that depend on it.

PR #153, opened 2020-06-23
Simplify entity change interface

dimagorelik:
Remove redundant operation from the interface.
Fix API method names

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/validators/EntityChangeValidator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/validators/EntityChangeValidator.java
@@ -1,6 +1,5 @@
 package com.kenshoo.pl.entity.internal.validators;
 
-import com.kenshoo.pl.entity.ChangeOperation;
 import com.kenshoo.pl.entity.Entity;
 import com.kenshoo.pl.entity.EntityChange;
 import com.kenshoo.pl.entity.EntityField;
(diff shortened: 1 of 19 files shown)


main/src/main/java/com/kenshoo/pl/entity/internal/validators/EntityChangeValidator.java
```diff
@@ -11,16 +10,36 @@
...
+     */
     SupportedChangeOperation getSupportedChangeOperation();
 
-    Stream<? extends EntityField<?, ?>> getFieldsToFetch(ChangeOperation changeOperation);
-
-    ValidationError validate(EntityChange<E> entityChange, Entity entity, ChangeOperation changeOperation);
+    /**
+     * @return a list of fields to fetch
+     */
+    Stream<? extends EntityField<?, ?>> fetchFields();
```
effiban, line 32: This name sounds like the method actually fetches fields.
I would rename to `fieldsToFetch()` or `fetchedFields()`

main/src/main/java/com/kenshoo/pl/entity/internal/validators/FieldsCombinationValidationAdapter.java
```diff
@@ -59,7 +53,7 @@ public ValidationError validate(EntityChange<E> entityChange, Entity entity, Cha
     }
 
     private boolean hasSubstitutions() {
-        return validator.substitutions().findAny() != null;
+        return validator.substitutions().findAny().isPresent();
```
effiban, line 56: :+1:

main/src/main/java/com/kenshoo/pl/entity/internal/validators/EntityChangeValidator.java
dimagorelik: @effiban 
Generally, I agree with you, but it is the property name in all our Interfaces.

dimagorelik: @effiban 
Method name - fetchFields

effiban: OK I see it is really in many other places, unfortunately :(

PR #155, opened 2020-06-30
PL-84: Replace FieldAuditRecord ctor. with builder

effiban:
Replacing ctor. with builder so that only one of the values can be passed and we don't need to pass any `null`s


main/src/main/java/com/kenshoo/pl/entity/audit/FieldAuditRecord.java
```diff
@@ -26,12 +28,16 @@ public FieldAuditRecord(final EntityField<E, ?> field,
...
     }
 
-    public Object getOldValue() {
-        return oldValue;
+    public Optional<?> getOldValue() {
+        return Optional.ofNullable(oldValue);
+    }
+
+    public Optional<?> getNewValue() {
+        return Optional.ofNullable(newValue);
```
effiban, line 36: These two are not directly related to the PR - but I wanted to fix this too. 
Originally I thought this class would be serialized directly to JSON but since it's not (the publisher will handle this) there is no reason to work with `null` directly.

effiban:
@amirz - FYI

PR #157, opened 2020-07-01
Added exception declaration for ChangeEntityCommand when using get method

michaelozerikenshoo:


main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
```diff
@@ -120,7 +120,7 @@ public boolean isFieldChanged(EntityField<E, ?> field) {
     }
 
     @Override
-    public <T> T get(EntityField<E, T> field) {
+    public <T> T get(EntityField<E, T> field) throws IllegalArgumentException {
```
effiban, line 123: What is this for?

effiban: @michaelozerikenshoo - the proper way to indicate a runtime exception like this one is through Javadoc, so if that's what you want to do - you are welcome to add it at the interface level (we really are missing Javadoc).
In any case, I am working on adding a new alternative method that will not throw any exceptions.

michaelozerikenshoo: Sorry this is a runtime error. thought declaring will add a compilation error since we have a lot of issues of excetopns in runtime when use used get before checking isFieldChanged. but we need to change the exception type.

PR #158, opened 2020-07-04
PL-87: Extract Triptionals.firstFilled() from EntityIdExtractor

effiban:
Following @galkoren -s suggestion: Extracting a static method from `EntityIdExtractor`, for finding the first 'filled' value from a sequence of `Triptional` suppliers. 
This method will be reused in the next PR where I will generalize `EntityIdExtractor` to support any field.
And then _finally_  I will be able to complete the task of supporting self-mandatory audited fields.

**General note:** Both `Triptional` and also the generalized extractor of the next PR - can be used elsewhere in the code to simplify it and avoid potential bugs.


effiban:
@amirz - FYI

PR #160, opened 2020-07-05
PL-88: Use FinalEntityState in EntityIdExtractor

effiban:
Refactoring `EntityIdExtractor` to use the new `FinalEntityState`.
I also replaced `Triptionals.isFilled()` with `Triptionals.isPresent()` to be more generic (there will be cases where we will want to handle the `null` value as a valid value)

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/EntityIdExtractor.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/EntityIdExtractor.java
@@ -4,7 +4,7 @@ import com.kenshoo.pl.entity.*;
 
 import java.util.Optional;
 
-import static com.kenshoo.pl.entity.internal.Triptionals.firstFilled;
+import static com.kenshoo.pl.entity.internal.Triptionals.firstPresent;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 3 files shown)


effiban:
@amirz - FYI

main/src/main/java/com/kenshoo/pl/entity/internal/EntityIdExtractor.java
```diff
@@ -25,16 +25,15 @@
...
+        return firstPresent(() -> extractFromIdentifier(entityChange, idField),
+                            () -> new FinalEntityState(currentState, entityChange).safeGet(idField))
             .mapToOptional(String::valueOf);
     }
 
     private <E extends EntityType<E>, T> Triptional<T> extractFromIdentifier(final EntityChange<E> entityChange,
                                                                              final EntityField<E, T> idField) {
         return Triptional.of(entityChange.getIdentifier())
-                         .flatMap(identifier -> identifier.safeGet(idField));
+                         .flatMap(identifier -> identifier.safeGet(idField), Triptional::absent);
```
effiban, line 36: The second param is to convert `null` into 'absent', because a missing identifier in this case is considered to be the same thing as a `null` identifier

PR #161, opened 2020-07-06
rename entity -> currentState

galkoren:


main/src/main/java/com/kenshoo/pl/entity/ChangeContext.java
```diff
@@ -10,7 +10,7 @@
 
     Entity getEntity(EntityChange entityChange);
 
-    void addEntity(EntityChange change, Entity entity);
+    void addEntity(EntityChange change, Entity currentState);
```
dimagorelik, line 13: @galkoren 
We need also rename the class Entity to CurrentState

galkoren: yes, next PR

PR #162, opened 2020-07-07
Rename EntityImpl -> CurrentEntityState

galkoren:
This is a preparation for FinalEntityState.
We shall end up with:
* CurrentEntityState implements Entity
* FinalEntityState implements Entity

Existing usages of the `Entity` interface are replaced with `CurrentEntityState` because Entity used to be the current state.
Those usages could work with "Entity", but it would be confusing.

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
@@ -81,7 +81,7 @@ abstract public class ChangeEntityCommand<E extends EntityType<E>> implements Mu
             }
 
             @Override
-            public T supply(Entity currentState) throws ValidationException {
+            public T supply(CurrentEntityState currentState) throws ValidationException {
                 FieldsValueMap<E> result = delegatingSupplier.supply(currentState);
                 if (result.containsField(entityField)) {
                     return result.get(entityField);
@@ -149,7 +149,7 @@ abstract public class ChangeEntityCommand<E extends EntityType<E>> implements Mu
         values.remove(field);
     }
 
-    void resolveSuppliers(Entity currentState) throws ValidationException {
+    void resolveSuppliers(CurrentEntityState currentState) throws ValidationException {
         // HashMap creates keySet/entrySet on demand so if the map is empty, calling these method implicitly increases its
         // memory consumption. Since in many cases the suppliers HashMap is empty, we can short-circuit this.
         if (suppliers.isEmpty()) {
--- a/main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
@@ -1,4 +1,4 @@
-package com.kenshoo.pl.entity.internal;
+package com.kenshoo.pl.entity;
 
 import com.kenshoo.pl.entity.Entity;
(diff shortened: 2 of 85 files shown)


main/src/main/java/com/kenshoo/pl/entity/CurrentEntityState.java
```diff
@@ -11,7 +11,19 @@
 import static java.util.Collections.emptyList;
 
 
-public class EntityImpl implements Entity {
+public class CurrentEntityState implements Entity {
+
+    public final  static CurrentEntityState EMPTY = new CurrentEntityState() {
```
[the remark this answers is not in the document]
galkoren: Because every method that used to accept an Entity, now accepts a CurrentEntityState.
And we cannot assign the `Entity.EMPTY` instance into a `CurrentEntityState` type.

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
```diff
@@ -36,13 +36,13 @@ public EntitiesFetcher(DSLContext dslContext, FeatureSet features) {
...
     }
 
-    public <E extends EntityType<E>> Map<Identifier<E>, Entity> fetchEntitiesByIds(final Collection<? extends Identifier<E>> ids,
-                                                                                   final EntityField<?, ?>... fieldsToFetchArgs) {
+    public <E extends EntityType<E>> Map<Identifier<E>, CurrentEntityState> fetchEntitiesByIds(final Collection<? extends Identifier<E>> ids,
+                                                                                               final EntityField<?, ?>... fieldsToFetchArgs) {
         return fetchEntitiesByIds(ids, ImmutableList.copyOf(fieldsToFetchArgs));
     }
 
-    public <E extends EntityType<E>> Map<Identifier<E>, Entity> fetchEntitiesByIds(final Collection<? extends Identifier<E>> ids,
```
dimagorelik, line 44: @galkoren 
Maybe better to return the Entity interface here?

main/src/main/java/com/kenshoo/pl/entity/ChangeEntityCommand.java
```diff
@@ -81,7 +81,7 @@ private void addCurrentStateConsumer(FetchEntityFields valueSupplier) {
             }
 
             @Override
-            public T supply(Entity currentState) throws ValidationException {
+            public T supply(CurrentEntityState currentState) throws ValidationException {
```
dimagorelik, line 84: @galkoren 
I have a problem with change. The Entity interface is immutable, but CurrentEntityState class is mutable. Why should we replace the Entity interface by  CurrentEntityState implementation?

galkoren: Good point 👍 
I think I can make the state (EntityImpl/CurrentEntityState) immutable.

dimagorelik: @galkoren 
Maybe to rename Entity - > CurrentEntityState

galkoren: I added a new class CurrentEntityMutableState to be what EntityImpl used to be.
CurrentEntityState is only a "marker" interface extending from Entity to prevent us from assigning a FinalState into CurrentState and via versa.

PR #163, opened 2020-07-13
add FinalEntityState to context

galkoren:


main/src/main/java/com/kenshoo/pl/entity/FinalEntityState.java
```diff
@@ -0,0 +1,32 @@
+package com.kenshoo.pl.entity;
+
+import java.util.List;
+
+public class FinalEntityState implements Entity {
+
+    final CurrentEntityState currentState;
+    final EntityChange<? extends EntityType<?>> change;
```
effiban, line 8: The above 2 should be `private`

main/src/main/java/com/kenshoo/pl/entity/FinalEntityState.java
```diff
@@ -0,0 +1,32 @@
+package com.kenshoo.pl.entity;
+
+import java.util.List;
+
+public class FinalEntityState implements Entity {
```
dimagorelik, line 5: @galkoren 
We already have a similar class. Please, see ResultingFieldsCombination

effiban: @dimagorelik - there are significant differences between them:
1) `ResultingFieldsCombination` is only relevant for fields of one entity type (same type as the command) but we need to support the more general case of fields from anywhere in the hierarchy
2) `ResultingFieldsCombination` filters by a stream of fields in the constructor and we don't want that limitation either
3) `ResultingFieldsCombination` implements `FieldsValueMap` and not `Entity` so it doesn't match the pattern we want of current entity and final entity implementing the same interface.

So it seems like `ResultingFieldsCombination` is a special case of `FinalyEntityState` (more or less) and I think we can leave it for a separate PR to see if and how we can combine them.

galkoren: I can't use it. The "Field Combination" is for a set of fields of a virtual field.
I don't have a set of fields.
The FinalState will soon calculate the complicated merging of fetchMany, which is not something the "Field Combination" could do.

dimagorelik: @galkoren 
I agree but maybe the new final state class can simplify the validator interface

effiban:
@galkoren - please add unit tests

PR #165, opened 2020-07-14
Fix Entity hierarchy generics

effiban:
- Add missing generic arguments (including '?' if needed)
- Add `@SuppressWarnings` to hide compiler warnings when casting is necessary


main/src/main/java/com/kenshoo/pl/entity/FinalEntityState.java
```diff
@@ -2,25 +2,26 @@
 
 import java.util.List;
 
-public class FinalEntityState implements Entity {
+public class FinalEntityState<E extends EntityType<E>> implements Entity {
```
dimagorelik, line 5: @effiban 
It shouldn' have generic type E because it implements Entity interface, but I agree that something is confusing in this interface.

effiban: The reason I added it is because that's the only way to avoid extra warnings and only do the minimal casting we need.

The reason this happens is because the `EntityChange::contains` and `EntityChange::get` method both _require_ the 'E' type, even though they can return a sensible result for any type (if the type doesn't match the answer should be `false`  or null/exception). 
On the other hand in the Java `Map` class (for example), the `get` method receives `Object` (and not 'K') for maximum flexibility.
I explained this to @galkoren but he doesn't want to change it since he likes the extra type-checking - plus anyway it would be a large refactor. So that's ok, but in order to avoid even more unnecessary warnings I prefer to add this 'E' here. Notice the 'E only on the class and not the interface, so the interface is still flexible as before.

effiban: @galkoren - I know we went back on forth on this when we did it together, after looking again I think this is the cleanest approach

effiban: I'm trying to follow best practices of Java, although I admit we have a tricky case here.
See for example this article:
https://www.informit.com/articles/article.aspx?p=2861454&seqNum=2

effiban:
@amirz @mmarusich - FYI

PR #166, opened 2020-07-15
EntitiesFetcher- Support fetching PartialEntity

libbyf:
Coverage tests already exists (EntitiesFetcherTest, PersistenceLayerTest)- checked in my local with FetchMany feature on


main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
```diff
@@ -110,8 +110,33 @@ public EntitiesFetcher(DSLContext dslContext, FeatureSet features) {
...
 
+    public <E extends EntityType<E>, PE extends PartialEntity, ID extends Identifier<E>> Map<ID, PE> fetchPartialEntities(E entityType, Collection<ID> ids, final Class<PE> entityIface) {
+        if (!features.isEnabled(FetchMany)) {
+            return oldEntityFetcher.fetchPartialEntities(entityType, ids, entityIface);
         }
+
+        final Map<Method, EntityField<E, ?>> entityMethodsMap = EntityTypeReflectionUtil.getMethodsMap(entityType, entityIface);
+        Map<Identifier<E>, CurrentEntityState> entitiesMap = fetchEntitiesByIds(ids, entityMethodsMap.values());
+
+        return seq(ids).filter(entitiesMap::containsKey).collect(toMap(identity(), id -> createInstance(entityIface, entityMethodsMap, entitiesMap.get(id))));
```
galkoren, line 124: Guava has a method to covert values of a map into a "view".
A "view" means that it doesn't copy the original map. It is a live transformation, so it is more efficient.
Do like this:
```java
return Maps.transformValues(entities, createInstance(ifc, methods, entities));
```

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
```diff
@@ -110,8 +110,33 @@ public EntitiesFetcher(DSLContext dslContext, FeatureSet features) {
...
+
+    public <E extends EntityType<E>, PE extends PartialEntity> List<PE> fetchByCondition(E entityType, final Condition condition, final Class<PE> entityIface) {
+        if (!features.isEnabled(FetchMany)) {
+            return oldEntityFetcher.fetchByCondition(entityType, condition, entityIface);
+        }
+
+        final Map<Method, EntityField<E, ?>> entityMethodsMap = EntityTypeReflectionUtil.getMethodsMap(entityType, entityIface);
+        final AliasedKey<?> aliasedKey = new AliasedKey<>(entityType.getPrimaryKey());
+
+        ArrayList<CurrentEntityState> entities = Lists.newArrayList(fetchEntities(entityType.getPrimaryTable(), aliasedKey, entityMethodsMap.values(), query -> query.withCondition(condition)).values());
```
galkoren, line 135: Don't need to convert Collection to List because you are steaming it on the next line

libbyf: :+1:

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesFetcher.java
libbyf: :+1:  good to know!!

PR #169, opened 2020-07-17
PL-74: Auditing: Support mandatory fields from same entity - part 2: Populate in AuditRecord

effiban:
See [part 1](https://github.com/kenshoo/persistence-layer/pull/148).

This part completes the feature by taking the "self-mandatory" fields that were resolved from annotations, and populating them in the resulting AuditRecord for a given command.

This means the following two things:
- Whenever a field is marked as `@Audited(trigger = ALWAYS)` and there is **any** change in the entity, it will appear in the top level of the generated record with the latest value.
- Whenever a field is marked as `@Audited(trigger = ALWAYS)` and the **field itself has changed**, it will appear **additionally** in the field records with the old and new values (like other fields).

So whenever the entity is created, and also on update if the mandatory field is changed - it will appear twice overall in the record.
This seeming duplication will help to keep the structure of the published record consistent and easy for parsing / analyzing.

**For example:**  

If the keyword entity is defined like this:

```
@Audited
public class KeywordType extends AbstractEntityType<KeywordType> {

    @Id
    public static final EntityField<Long> ID = INSTANCE.field(KeywordTable.INSTANCE.id);

    @Audited(trigger = ALWAYS)
    public static final EntityField<KeywordType, String> MATCH_TYPE = INSTANCE.field(KeywordTable.INSTANCE.match_type);

    public static final EntityField<KeywordType, String> TEXT = INSTANCE.field(KeywordTable.INSTANCE.text);
}
```

Then after creating a new keyword the audit record might look like this:

```
{
    "entityType": "Keyword",
    "entityId: 1234,
    "matchType": "BROAD",
    "operator": "CREATE",
    "fieldChanges": [
         "matchType" : {
               "newValue" : "BROAD"
         },
         "text": {
               "newValue" : "Shoes"
         }
    ]
}
```

After updating the text it may look like this:

```
{
    "entityType": "Keyword",
    "entityId: 1234,
    "matchType": "BROAD",
    "operator": "UPDATE",
    "fieldChanges": [
         "text": {
               "oldValue" : "Shoes",
               "newValue" : "Pants"
         }
    ]
}
```

After updating the match type it may look like this:

```
{
    "entityType": "Keyword",
    "entityId: 1234,
    "matchType": "PHRASE",
    "operator": "UPDATE",
    "fieldChanges": [
         "matchType": {
               "oldValue" : "BROAD",
               "newValue" : "PHRASE"
         }
    ]
}
```

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
@@ -65,7 +65,7 @@ public class AuditRecordGenerator<E extends EntityType<E>> implements CurrentSta
 
         final Collection<? extends EntityFieldValue> mandatoryFieldValues = generateMandatoryFieldValues(finalState);
 
-        final Collection<? extends FieldAuditRecord<E>> fieldRecords = generateFieldRecords(currentState, finalState);
+        final Collection<? extends FieldAuditRecord<E>> fieldRecords = generateChangedFieldRecords(currentState, finalState);
 
         return new AuditRecord.Builder<E>()
             .withEntityType(entityChange.getEntityType())
@@ -77,7 +77,7 @@ public class AuditRecordGenerator<E extends EntityType<E>> implements CurrentSta
             .build();
     }
 
-    private Collection<? extends FieldAuditRecord<E>> generateFieldRecords(final CurrentEntityState currentState, final FinalEntityState finalState) {
+    private Collection<? extends FieldAuditRecord<E>> generateChangedFieldRecords(final CurrentEntityState currentState, final FinalEntityState finalState) {
         return auditedFieldSet.getAllSelfFields()
                               .filter(field -> fieldWasChanged(currentState, finalState, field))
                               .map(field -> buildFieldRecord(currentState, finalState, field))
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
@@ -10,7 +10,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 
 import java.util.Collection;
 import java.util.Optional;
-import java.util.Set;
 import java.util.stream.Stream;
 
 import static com.kenshoo.pl.entity.ChangeOperation.UPDATE;
@@ -62,14 +61,11 @@ public class AuditRecordGenerator<E extends EntityType<E>> implements CurrentSta
 
         final String entityId = extractEntityId(entityChange, currentState);
 
-        final Collection<? extends EntityFieldValue> mandatoryFieldValues = generateMandatoryFieldValues(currentState);
+        final FinalEntityState finalState = new FinalEntityState(currentState, entityChange);
 
-        final Set<? extends EntityField<E, ?>> candidateOnChangeFields = auditedFieldSet.intersectWith(entityChange.getChangedFields())
-                                                                                        .getOnChangeFields();
+        final Collection<? extends EntityFieldValue> mandatoryFieldValues = generateMandatoryFieldValues(finalState);
 
-        final Collection<? extends FieldAuditRecord<E>> fieldRecords = generateFieldRecords(entityChange,
-                                                                                            currentState,
-                                                                                            candidateOnChangeFields);
+        final Collection<? extends FieldAuditRecord<E>> fieldRecords = generateFieldRecords(currentState, finalState);
 
         return new AuditRecord.Builder<E>()
             .withEntityType(entityChange.getEntityType())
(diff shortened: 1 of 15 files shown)


effiban:
@amirz @libbyf - FYI

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
```diff
@@ -48,8 +48,14 @@ private AuditedFieldSet(final EntityField<E, ? extends Number> idField,
         return selfMandatoryFields;
     }
 
-    public Set<? extends EntityField<E, ?>> getOnChangeFields() {
-        return onChangeFields;
+    public Stream<? extends EntityField<?, ?>> getAllMandatoryFields() {
```
galkoren, line 51: why "All"? Is there a risk that `getMandatoryFields` will miss out some fields?

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
```diff
@@ -81,21 +77,19 @@ public AuditRecordGenerator(final AuditedFieldSet<E> auditedFieldSet) {
...
-    private Collection<? extends FieldAuditRecord<E>> generateFieldRecords(final EntityChange<E> entityChange,
-                                                                           final CurrentEntityState currentState,
-                                                                           final Collection<? extends EntityField<E, ?>> candidateOnChangeFields) {
-        return candidateOnChangeFields.stream()
-                                      .filter(field -> fieldWasChanged(entityChange, currentState, field))
-                                      .map(field -> buildFieldRecord(entityChange, currentState, field))
-                                      .collect(toList());
+    private Collection<? extends FieldAuditRecord<E>> generateFieldRecords(final CurrentEntityState currentState, final FinalEntityState finalState) {
+        return auditedFieldSet.getAllSelfFields()
+                              .filter(field -> fieldWasChanged(currentState, finalState, field))
```
galkoren, line 82: I prefer the previous logic. Why do we re-calculate the EntityChange as a diff of Final and Current?

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
effiban: "All" means both external and self mandatory - this makes it clear what the method returns.
We need it because both the self and external mandatory are placed in the same section of the AuditRecord (top level of the JSON with the latest values).
In addition there are separate getters for the external and the self in case they are needed (although not at the moment and may be removed later).

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
effiban: I tried a few different implementations here and I think this is the cleanest.
By calculating it in this way it covers all the conditions you need to check, otherwise at least one additional "if" is needed as far as I can tell.
This way you get a fully _functional-style_ implementation that uses only the `safeGet()`, and will allow us to get rid of both `containsField` and `get()` later on.
I think it's also more readable and reduces the chance for bugs

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
```diff
@@ -81,21 +77,19 @@ public AuditRecordGenerator(final AuditedFieldSet<E> auditedFieldSet) {
...
     }
 
-    private Collection<? extends FieldAuditRecord<E>> generateFieldRecords(final EntityChange<E> entityChange,
-                                                                           final CurrentEntityState currentState,
-                                                                           final Collection<? extends EntityField<E, ?>> candidateOnChangeFields) {
-        return candidateOnChangeFields.stream()
-                                      .filter(field -> fieldWasChanged(entityChange, currentState, field))
-                                      .map(field -> buildFieldRecord(entityChange, currentState, field))
-                                      .collect(toList());
+    private Collection<? extends FieldAuditRecord<E>> generateFieldRecords(final CurrentEntityState currentState, final FinalEntityState finalState) {
```
galkoren, line 80: generateChange(Fields)Records

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
```diff
@@ -106,30 +100,24 @@ private String extractEntityId(final EntityChange<E> entityChange,
...
     }
 
-    private <T> boolean fieldValuesEqual(final EntityChange<E> entityChange,
-                                         final CurrentEntityState currentState,
-                                         final EntityField<E, T> field) {
-        return field.valuesEqual(entityChange.get(field),  currentState.get(field));
+    private <T> boolean fieldStayedTheSame(final CurrentEntityState currentState,
+                                           final FinalEntityState finalState,
+                                           final EntityField<E, T> field) {
+        return currentState.safeGet(field).equals(finalState.safeGet(field), field::valuesEqual);
```
galkoren, line 120: Triptional shall pass `null` to the user function

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGenerator.java
effiban: I will rename the method, however it generates an inconsistency with other places in the code where they are called just "fieldRecords". 
I would rather not do a full refactoring now but we can discuss it separately if you want.

effiban:
@galkoren - comments fixed

PR #172, opened 2020-07-21
PL-65: Support auditing on update only - part 1: Add trigger type

effiban:
Adding support for a new audit trigger rule: Audit a field only if it has been changed by an update, and not upon create.

This is part 1 handling the trigger enums only: 
- Renamed **ON_CHANGE** to **ON_CREATE_OR_UPDATE** 
- Added **ON_UPDATE** which means to audit only if changed by an update.

In part 2, the new annotation will be resolved and placed in a new member in `AuditedFieldSet`

_Note_: This feature is required by the Kenshoo Search application

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/annotation/audit/Audited.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/annotation/audit/Audited.java
@@ -9,7 +9,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-import static com.kenshoo.pl.entity.audit.AuditTrigger.ON_CHANGE;
+import static com.kenshoo.pl.entity.audit.AuditTrigger.ON_CREATE_OR_UPDATE;
 
 /**
  * Whenever an entity or field has this annotation, it indicates that any changes to the entity / field
@@ -22,9 +22,10 @@ public @interface Audited {
 
     /**
      * @return the rule by which to trigger auditing for the annotated entity type or field.<br>
-     * This attribute is valid for <b>field-level annotations only</b>, and will be ignored if appearing on entities (for the entity-level, ON_CHANGE is implied always).
+     * This attribute is valid for <b>field-level annotations only</b>, and will be ignored if appearing on entities.<br>
+     * For the entity-level,{@link AuditTrigger#ON_CREATE_OR_UPDATE} is implied always.
      */
-    AuditTrigger trigger() default ON_CHANGE;
+    AuditTrigger trigger() default ON_CREATE_OR_UPDATE;
 
     /**
      * <b>NOTE</b>: This attribute is valid for entity-level annotations only, and will be ignored if appearing on fields.
--- a/main/src/main/java/com/kenshoo/pl/entity/audit/AuditTrigger.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/audit/AuditTrigger.java
@@ -8,13 +8,22 @@ package com.kenshoo.pl.entity.audit;
 public enum AuditTrigger {
 
     /**
-     * Indicates that a field should be audited always, regardless of whether its value has changed - whenever there is some other change in the  currentState.<br>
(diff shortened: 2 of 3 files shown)


effiban:
@libbyf @mmarusich @amirz  - FYI

main/src/main/java/com/kenshoo/pl/entity/annotation/audit/Audited.java
```diff
@@ -22,9 +22,10 @@
 
     /**
      * @return the rule by which to trigger auditing for the annotated entity type or field.<br>
-     * This attribute is valid for <b>field-level annotations only</b>, and will be ignored if appearing on entities (for the entity-level, ON_CHANGE is implied always).
+     * This attribute is valid for <b>field-level annotations only</b>, and will be ignored if appearing on entities.<br>
+     * For the entity-level,{@link AuditTrigger#ON_CREATE_OR_UPDATE} is implied always.
      */
-    AuditTrigger trigger() default ON_CHANGE;
+    AuditTrigger trigger() default ON_CREATE_OR_UPDATE;
```
dimagorelik, line 28: @effiban 
I think that the default should be ON_UPDATE

main/src/main/java/com/kenshoo/pl/entity/audit/AuditTrigger.java
```diff
@@ -8,13 +8,22 @@
 public enum AuditTrigger {
```
dimagorelik, line 8: @effiban 
Can we use this one instead SupportedChangeOperation?

main/src/main/java/com/kenshoo/pl/entity/annotation/audit/Audited.java
effiban: @galkoren - WDYT?

main/src/main/java/com/kenshoo/pl/entity/audit/AuditTrigger.java
effiban: They are similar but don't have the same meaning exactly. 
Here we have the option of "ALWAYS" which means to include the field always in a particular section even if it wasn't changed at all. Even though we could make it work, the meaning of the individual enums would not be clear enough (I think). So the javadoc I added here is important

main/src/main/java/com/kenshoo/pl/entity/annotation/audit/Audited.java
effiban: @mmarusich - WDYT? What would be a sensible default for Social?

mmarusich: @effiban I think default should be on_create_update as we need history that entity was created, we do display this in our UX

effiban: @mmarusich - I agree, just want to point out that there will always be a history of creation at **entity** level (even in KS).
The question is whether you need to have all the fields audited upon create or only a few representative fields like name , and then all others will appear only if they are updated.

mmarusich: I suppose we need all, it make sense for me as you will want to get info from history regarding initial state of entity

dimagorelik: @mmarusich @effiban 
It will cause services to send a big amount of data and will require huge storage (Snowflake, Kafka). I do not think that we need to duplicate the data in the Khystory service. The created data is stored in the original tables. I think sending a few representative fields like the name for create operation is more reasonable.

dimagorelik: @effiban 
By the way, what is about delete operation? Should we also send all data?

effiban: @dimagorelik - for now, we are not sending anything on delete except for the mandatory fields

PR #173, opened 2020-07-23
Fix any change trigger in composite validator

dimagorelik:
Fix any change trigger in composite validator


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidator.java
```diff
@@ -20,6 +20,7 @@
 
 public class EntityChangeCompositeValidator<E extends EntityType<E>> implements ChangesValidator<E> {
 
+    private final AnyFieldTrigger<E> ANY_FIELD_TRIGGER = new AnyFieldTrigger<>();
```
effiban, line 23: If you want to make it a singleton - then please do it in the conventional way: 
Define a `public static final` method in `AnyFieldTrigger` to return the instance + a private constructor

dimagorelik: @effiban 
I do not want to do it a singleton. I want one instance per EntityChangeCompositeValidator.

effiban: Why not a singleton? There are no members in the validator

PR #174, opened 2020-07-25
PL-93: Extract AuditedFieldSet.intersectWith() to AuditedFieldsToFetchResolver

effiban:
The method `AuditedFieldSet.intersectWith()` is currently used in only one place - for calculating the fields to fetch in `AuditRecordGenerator.requiredFields()`.
It will soon get more complicated when I add the logic for the ON_UPDATE fields.
So instead of keeping this complicated method in a POJO, with all of the associated tests - I'm extracting it to its own class `AuditedFieldsToFetchResolver`


effiban:
@amirz - FYI

PR #175, opened 2020-07-26
PL-90: Support auditing on update only, part 2: Resolve annotation

effiban:
See [part 1](https://github.com/kenshoo/persistence-layer/pull/172).
In this part, resolving the `ON_UPDATE` annotation from entity fields and placing them in the `AuditedFieldSet` which now contains a multimap of all the "internal" (formerly called "self") fields, keyed by trigger type.

At this point the `ON_UPDATE` annotated fields are still not generated according to the rule in the `AuditRecord` (that will be handled in the next part).


effiban:
@amirz  - FYI

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
```diff
@@ -81,7 +94,8 @@ public boolean hasSelfFields() {
         private final EntityField<E, ? extends Number> idField;
         private Set<? extends EntityField<?, ?>> externalMandatoryFields = emptySet();
         private Set<? extends EntityField<E, ?>> selfMandatoryFields = emptySet();
-        private Set<? extends EntityField<E, ?>> onChangeFields = emptySet();
+        private Set<? extends EntityField<E, ?>> onCreateOrUpdateFields = emptySet();
```
galkoren, line 97:
```
List<field> external;
MultiMap<TriggerType, Field> internal;
```

effiban:
@galkoren - I replaced the separate sets with a multimap as requested - it caused a large refactor. Please review again

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditedFieldSet.java
```diff
@@ -18,59 +24,52 @@
...
+
+    // Fields from current entity (besides id), keyed by audit trigger type
+    private final SetMultimap<AuditTrigger, ? extends EntityField<E, ?>> internalFields;
 
     private AuditedFieldSet(final EntityField<E, ? extends Number> idField,
-                            final Set<? extends EntityField<?, ?>> externalMandatoryFields,
-                            final Set<? extends EntityField<E, ?>> selfMandatoryFields,
-                            final Set<? extends EntityField<E, ?>> onChangeFields) {
+                            final Set<? extends EntityField<?, ?>> externalFields,
+                            final SetMultimap<AuditTrigger, ? extends EntityField<E, ?>> internalFields) {
```
galkoren, line 36: Why SetMultimap and not just any Multimap?

effiban: Because there shouldn't be any duplicates within each type - just like the individual `Set`-s that were there before.

PR #177, opened 2020-08-05
PL-95: Extract interfaces for AuditRecordGenerator

effiban:
Since more logic needs to be added and the class is already complicated - I'm starting a series of refactorings to break up `AuditRecordGenerator` into smaller parts.
The first step is here as follows:
1) Renamed `AuditRecordGenerator` class to `AuditRecordGeneratorStateConsumerImpl` 
2) Extracted an interface `AuditRecordGeneratorStateConsumer` for the whole class
3) Extracted an interface `AuditRecordGenerator` for the `generate()` method only, and made the interface above extend it

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
@@ -4,7 +4,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.kenshoo.pl.entity.internal.*;
-import com.kenshoo.pl.entity.internal.audit.AuditRecordGenerator;
+import com.kenshoo.pl.entity.internal.audit.AuditRecordGeneratorStateConsumer;
+import com.kenshoo.pl.entity.internal.audit.AuditRecordGeneratorStateConsumerImpl;
 import com.kenshoo.pl.entity.internal.audit.AuditedFieldsResolver;
 import com.kenshoo.pl.entity.spi.*;
 import com.kenshoo.pl.entity.spi.helpers.EntityChangeCompositeValidator;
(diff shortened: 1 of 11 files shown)


main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -35,7 +35,7 @@
     private final List<ChangesFilter<E>> postFetchFilters;
     private final List<ChangesFilter<E>> postSupplyFilters;
     private final PersistenceLayerRetryer retryer;
-    private final AuditRecordGenerator<E> auditRecordGenerator;
+    private final AuditRecordGeneratorImpl<E> auditRecordGenerator;
```
effiban, line 38: Here we must use the impl and not the interface, because it also has the `CurrentStateConsumer` implementation in it

effiban:
@amirz - FYI

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
dimagorelik, line 38: @effiban 
Could AuditRecordGenerator extend CurrentStateConsumer?

effiban: I don't want it to, because in the next refactor I will break the impl into 2 parts, and the new interface is only for the **second** part.
I can also create an interface that will have everything, but there is no specific usage for it

effiban: My plan is to rename the current impl to **AuditRecordGeneratorStateConsumer** and that's what will appear here

effiban: The name with "Impl" at the end will be used for the generation logic only

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -47,7 +47,7 @@ private ChangeFlowConfig(E entityType,
                              Set<EntityField<E, ?>> requiredFields,
                              List<ChangeFlowConfig<? extends EntityType<?>>> childFlows,
                              PersistenceLayerRetryer retryer,
-                             AuditRecordGenerator<E> auditRecordGenerator,
+                             AuditRecordGeneratorImpl<E> auditRecordGenerator,
```
dimagorelik, line 50: @effiban 
I think it is bettrer that Ctr will receive interface

PR #180, opened 2020-08-09
PL-98: Split AuditRecordGenerator

effiban:
Splitting `AuditRecordGeneratorImpl` into smaller classes handling specific groups of fields.
The motivation for doing this is:
1) Following the single-responsibility principle :)
2) Reduce and simplify the unit tests of each class so they are easier to understand and maintain
3) Prepare for the next refactoring which will implement the trigger of auditing "on update" only

See diagrams:

**BEFORE:**

image

**AFTER:**

image

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
@@ -283,7 +283,7 @@ public class ChangeFlowConfig<E extends EntityType<E>> {
             final AuditRequiredFieldsCalculator<E> auditRequiredFieldsCalculator =
                 optionalAuditedFieldSet.map(AuditRequiredFieldsCalculator::new).orElse(null);
             final AuditRecordGenerator<E> auditRecordGenerator =
-                optionalAuditedFieldSet.map(AuditRecordGeneratorImpl::new).orElse(null);
+                optionalAuditedFieldSet.map(this::createAuditRecordGenerator).orElse(null);
 
             return new ChangeFlowConfig<>(entityType,
                                           enrichers.build(),
@@ -297,6 +297,15 @@ public class ChangeFlowConfig<E extends EntityType<E>> {
                                           auditRecordGenerator,
                                           features
             );
+        }
+
+        private AuditRecordGenerator<E> createAuditRecordGenerator(final AuditedFieldSet<E> auditedFieldSet) {
+            final AuditMandatoryFieldValuesGenerator mandatoryFieldValuesGenerator =
+                new AuditMandatoryFieldValuesGenerator(auditedFieldSet.getMandatoryFields());
+
+            final AuditFieldChangesGenerator<E> fieldChangesGenerator = new AuditFieldChangesGenerator<>(auditedFieldSet.getInternalFields());
+
+            return new AuditRecordGeneratorImpl<>(mandatoryFieldValuesGenerator, fieldChangesGenerator);
         }
 
         static private class Labeled<Element> {
--- a/main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
@@ -10,9 +10,7 @@ import java.util.Optional;
 
 import static java.util.Objects.requireNonNull;
 
-class AuditFieldChangeGenerator {
-
-    static final AuditFieldChangeGenerator INSTANCE = new AuditFieldChangeGenerator();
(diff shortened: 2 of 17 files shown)


main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
```diff
@@ -0,0 +1,49 @@
...
+import com.kenshoo.pl.entity.FinalEntityState;
+import com.kenshoo.pl.entity.audit.FieldAuditRecord;
+
+import java.util.Optional;
+
+import static java.util.Objects.requireNonNull;
+
+class AuditFieldChangeGenerator {
+
+    static final AuditFieldChangeGenerator INSTANCE = new AuditFieldChangeGenerator();
```
dimagorelik, line 15: @effiban 
Why should this class be a singleton?

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
```diff
@@ -0,0 +1,49 @@
...
+
+    <E extends EntityType<E>> Optional<FieldAuditRecord<E>> generate(final CurrentEntityState currentState,
```
dimagorelik, line 17: @effiban 
You can define the method as static instead of a singleton

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
effiban: @dimagorelik - why not? That is the usual pattern

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditFieldChangeGenerator.java
effiban: @dimagorelik - the issue is unit tests, the method can be mocked in the usual way only if it's a non-static method.
That's also why I did all of this refactoring in the first place - to make the tests more manageable (and also to prepare for the next refactor)

dimagorelik: @effiban 
There is no state in this class. 
We can simply to create a new instance

effiban: Not sure I understand - it just means we will be creating more than one instance unnecessarily.. that's why singleton is better

effiban:
@amirz - FYI

effiban:
@dimagorelik  - fixed your comment

main/src/main/java/com/kenshoo/pl/entity/internal/audit/AuditRecordGeneratorImpl.java
```diff
@@ -53,11 +61,11 @@ public AuditRecordGeneratorImpl(final AuditedFieldSet<E> auditedFieldSet) {
 
         final String entityId = extractEntityId(entityChange, currentState);
 
-        final FinalEntityState finalState = new FinalEntityState(currentState, entityChange);
+        final FinalEntityState finalState = finalStateCreator.apply(currentState, entityChange);
```
dimagorelik, line 64: @effiban 
Can we take the final state from context?

effiban: @dimagorelik - I checked and you're right, I thought we can't pass the context to the generator but in fact we have it, and now it makes sense to me.
I'll make this change.

effiban:
@dimagorelik - fixed second comment

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -299,6 +299,15 @@ public Builder(E entityType) {
...
         }
 
+        private AuditRecordGenerator<E> createAuditRecordGenerator(final AuditedFieldSet<E> auditedFieldSet) {
+            final AuditMandatoryFieldValuesGenerator mandatoryFieldValuesGenerator =
+                new AuditMandatoryFieldValuesGenerator(auditedFieldSet.getMandatoryFields());
+
+            final AuditFieldChangesGenerator<E> fieldChangesGenerator = new AuditFieldChangesGenerator<>(auditedFieldSet.getInternalFields());
+
+            return new AuditRecordGeneratorImpl<>(mandatoryFieldValuesGenerator, fieldChangesGenerator);
+        }
```
effiban, line 309: will be extracted to a builder later on

PR #182, opened 2020-08-17
PL-94: Java 11 upgrade

effiban:
It was easier than I expected...

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ PL precalculates the fields required by the flow components and fetches exactly 
 It also precalculates which validators are really required by your commands and there are more optimizations along the way. Oh, and it does not use reflection.
 
 ## Compatibility
-* Java 8 or greater
+* Java 11 or greater
 * MySQL Database (5.6 or greater)
 * PL depends on [JOOQ](https://www.jooq.org). Each PL release version specifies the JOOQ version it was built with. E.g. release "0.1.40-jooq-3.10.4" was built with JOOQ 3.10.4. If you need a very specific JOOQ version, we can add it to our build process.
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ PL precalculates the fields required by the flow components and fetches exactly 
 It also precalculates which validators are really required by your commands and there are more optimizations along the way. Oh, and it does not use reflection.
 
 ## Compatibility
-* Java 11 or greater
+* Java 11 or greater (version **0.1.51** or greater)
+* Java 8 (up to version **0.1.50**)
 * MySQL Database (5.6 or greater)
 * PL depends on [JOOQ](https://www.jooq.org). Each PL release version specifies the JOOQ version it was built with. E.g. release "0.1.40-jooq-3.10.4" was built with JOOQ 3.10.4. If you need a very specific JOOQ version, we can add it to our build process.
 
@@ -134,7 +135,7 @@ From maven
 <dependency>
     <groupId>com.kenshoo</groupId>
     <artifactId>persistence-layer</artifactId>
-    <version>0.1.50-jooq-3.10.4</version>
+    <version>0.1.51-jooq-3.10.4</version>
 </dependency>
 ```
 
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -23,6 +23,4 @@ project.ext.libraries = [
         guavaRetrying        : 'com.github.rholder:guava-retrying:2.0.0',
 
         dockerTest           : 'com.kenshoo:docker-test:4.5.+',
-
-        jaxbApi              : 'javax.xml.bind:jaxb-api:2.3.1'
 ]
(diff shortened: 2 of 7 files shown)


effiban:
@morantavori - FYI

dimagorelik:
@EB123 
Please, attach the list of warnings.

effiban:
@dimagorelik  - I fixed the first warning by adding **jaxb-api** which is a library that was extracted from the JDK in Java 11.
This was only a warning because it has to do with annotations in Jooq code that are not directly used by our code.

This is the second:

image

We have seen this in other Java 11 upgrades due to third-party libs we are using that are not fully compatible with Java 11 which introduced stricter access rules. 
However they are only warnings and won't cause errors until later JDK versions. I'm not sure where exactly this one comes from, will require some investigation but I'm not sure it's worth it right now.

effiban:
@dimagorelik I checked, and it seems to be coming from the Mockito version which depends on an old version of bytebuddy (bytecode manipulation lib). 
I will try to upgrade

dependencies.gradle
```diff
@@ -22,6 +22,7 @@ project.ext.libraries = [
 
         guavaRetrying        : 'com.github.rholder:guava-retrying:2.0.0',
 
-        dockerTest           : 'com.kenshoo:docker-test:4.5.+'
+        dockerTest           : 'com.kenshoo:docker-test:4.5.+',
 
+        jaxbApi              : 'javax.xml.bind:jaxb-api:2.3.1'
```
dimagorelik, line 27: @effiban 
Whare does PL use jaxApi? 
I do not think we need to add this dependency

main/build.gradle
```diff
@@ -11,10 +11,10 @@ dependencies {
     compile libraries.jool
     compile libraries.guava
 
-    compile libraries.jool
     compile libraries.apacheCommons3
     compile libraries.mysqlConnector
     compile libraries.guavaRetrying
+    compile libraries.jaxbApi
```
dimagorelik, line 17: @effiban 
Do we need it for production?

effiban: @dimagorelik - as explained above this was extracted from Java so anyone who used it before in production will need to add it as a separate dependency.
Most of the Kenshoo projects have it already (you can check **search** which has just been upgraded).

dimagorelik: @effiban 
Is it the same version?
Is it good for an open-source library?

effiban: The problem is actually more complicated, and apparently it only really causes issues with users of the jooq generation tool.
There are many posts about this, for example:
https://stackoverflow.com/questions/59690301/unable-to-run-jooq-generationtool-with-java-11-noclassdeffounderror-javax-xml/59703860#59703860

Maybe it has been resolved in later Jooq versions, but I couldn't figure that out from a quick Google search.
In any case, I don't see any major issue with adding it here.

effiban: Actually search has an older jaxb version and we will probably need to remove it from there, so it can get the version from here. 
This goes back to the question of the jooq version - I think we should only support one jooq version and force the clients to use it, and jaxb goes together with jooq.
Otherwise we will have too much trouble maintaining the versions.

effiban: We can also decide to shadow this lib so clients will not get it and our compilation will still be free of warnings. 
But I still think that forcing the clients to the same Jooq version is the right way to go. There's no good reason I can think of to hold back on Jooq upgrades, especially now that the Java version is aligned.

README.md
```diff
@@ -28,7 +28,7 @@ PL precalculates the fields required by the flow components and fetches exactly
 It also precalculates which validators are really required by your commands and there are more optimizations along the way. Oh, and it does not use reflection.
 
 ## Compatibility
-* Java 8 or greater
+* Java 11 or greater
```
galkoren, line 31: Let's write that Java 8 for version <= 51 , and Java 11 for later releases.

main/build.gradle
galkoren: JAXB is only required when generating tables using a command line, a feature used only by Social.
We can remove this dependency and just add a note about it in the wiki.
If JOOQ did not add this dependency explicitly, then neither should we. They (Lucas from JOOQ) are supposed to decide whether to include this dep or whether to fix their bug.

effiban: Ok - I'll remove it. Still I think it's strange that we will have the following warning always:
```
warning: unknown enum constant XmlAccessType.FIELD
  reason: class file for javax.xml.bind.annotation.XmlAccessType not found
```

README.md
effiban: I'm not sure this is necessary, since very soon nobody will be using the older versions, and we are not going to support two major versions (I hope).
I don't mind adding it though

effiban:
@galkoren @dimagorelik - all comments fixed

effiban:
@amirz @idopreuss  - FYI

PR #183, opened 2020-08-19
Trigger ancestors validator for empty changes

dimagorelik:
Trigger ancestors validator for empty changes


main/src/test/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidatorTest.java
```diff
@@ -184,4 +184,21 @@ public void registerAncestorsValidatorForUpdateTest() {
         validator.validate(entityChanges, ChangeOperation.UPDATE, changeContext);
         verify(ancestorsValidator).validate(any(CurrentEntityState.class));
     }
+
+    @Test
+    public void registerAncestorsValidatorForEmptyUpdateTest() {
+        when(entityChange.getChangedFields()).thenReturn(Stream.of());
+        validator.register(ancestorsValidator);
+        validator.validate(entityChanges, ChangeOperation.UPDATE, changeContext);
+        verify(ancestorsValidator).validate(any(CurrentEntityState.class));
```
effiban, line 193: The current state is available as a mock object, please use it instead calling `any()`

effiban (commented):
One minor comment regarding mockito.

Aside from that - I see you added tests for the problematic case of empty update.
Please check that the existing test coverage for other cases is good enough, since you made a significant change

main/src/test/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidatorTest.java
dimagorelik: @effiban 
fixed

effiban (approved):

PR #184, opened 2020-08-20
Entity change composite validator refactoring

dimagorelik:
Entity change composite validator refactoring fix the bug of ancestor validator when the changed entity does not contain any changed fields.

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidator.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidator.java
@@ -17,13 +17,11 @@ import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static java.util.stream.Collectors.toSet;
-
 public class EntityChangeCompositeValidator<E extends EntityType<E>> implements ChangesValidator<E> {
 
-    private final ValidationTrigger<E> ANY_CHANGE_TRIGGER = entityField -> true;
+    private final ValidationTrigger<E> ANY_CHANGE_TRIGGER = entityFields -> true;
 
-    private final List<Tuple2<ValidationTrigger<E>, EntityChangeValidator<E>>> changesValidators = new ArrayList<>();
+    private final List<Tuple2<ValidationTrigger<E>, EntityChangeValidator<E>>> triggeredChangeValidators = new ArrayList<>();
 
     public void register(E entityType, ChangeValidator validator) {
         if (validator instanceof FieldComplexValidator) {
@@ -75,11 +73,11 @@ public class EntityChangeCompositeValidator<E extends EntityType<E>> implements 
     }
 
     public void register(AncestorsValidator validator) {
-        changesValidators.add(Tuple.tuple(ANY_CHANGE_TRIGGER, new AncestorsValidationAdapter<>(validator)));
+        triggeredChangeValidators.add(Tuple.tuple(ANY_CHANGE_TRIGGER, new AncestorsValidationAdapter<>(validator)));
     }
 
     public void register(EntityChangeValidator<E> validator) {
-        changesValidators.add(Tuple.tuple(new AnyFieldsTrigger<>(validator.validatedFields()), validator));
+        triggeredChangeValidators.add(Tuple.tuple(new AnyFieldsTrigger<>(validator.validatedFields()), validator));
     }
 
     public <T> void register(E entityType, PrototypeFieldValidator<T> validator) {
@@ -88,7 +86,7 @@ public class EntityChangeCompositeValidator<E extends EntityType<E>> implements 
             throw new IllegalArgumentException("Can not find entity field by prototype: " + validator.getPrototype());
         }
         for (EntityField<E, T> entityField : entityFields) {
-            changesValidators.add(Tuple.tuple(new FieldTrigger<>(entityField), new PrototypeFieldValidationAdapter<>(entityField, validator)));
+            triggeredChangeValidators.add(Tuple.tuple(new FieldTrigger<>(entityField), new PrototypeFieldValidationAdapter<>(entityField, validator)));
         }
     }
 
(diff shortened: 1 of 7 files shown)


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidator.java
```diff
@@ -119,9 +119,9 @@ public void register(E entityType, PrototypeFieldsCombinationValidator validator
     public void validate(Collection<? extends EntityChange<E>> entityChanges, ChangeOperation changeOperation, ChangeContext changeContext) {
         entityChanges.forEach(entityChange -> {
             CurrentEntityState currentState = changeContext.getEntity(entityChange);
-            Collection<EntityChangeValidator<E>> validators = findValidators(entityChange, changeOperation);
+            Collection<? extends EntityField<E, ?>> fieldsToUpdate = entityChange.getChangedFields().collect(Collectors.toList());
+            Collection<EntityChangeValidator<E>> validators = findValidatorsTriggeredByFields(fieldsToUpdate, changeOperation).collect(toSet());
```
dimagorelik, line 123: @effiban @galkoren 
Is it better to use here distinct?

effiban: @dimagorelik - according to [this](https://stackoverflow.com/questions/41593538/is-it-worth-using-distinct-with-collecttoset) the answer is "yes"

main/src/test/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidatorTest.java
```diff
@@ -184,4 +184,22 @@ public void registerAncestorsValidatorForUpdateTest() {
...
+        validator.validate(entityChanges, ChangeOperation.UPDATE, changeContext);
+        verify(ancestorsValidator).validate(currentState);
+    }
+
+    @Test
+    public void ancestorsRequiredFieldsForEmptyUpdateTest() {
+        when(ancestorsValidator.ancestorsFields()).thenReturn(Stream.of(TestEntity.FIELD_1));
+        validator.register(ancestorsValidator);
+        List<? extends EntityField<?, ?>> fields = validator.requiredFields(Collections.emptyList(), ChangeOperation.UPDATE).collect(Collectors.toList());
+        ;
```
effiban, line 201: remove redundant `;`

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidator.java
```diff
@@ -130,22 +130,14 @@ public void validate(Collection<? extends EntityChange<E>> entityChanges, Change
...
-                .collect(toSet());
-    }
-
-    private Stream<EntityChangeValidator<E>> findValidatorsTriggeredByField(EntityField<E, ?> entityField) {
-        return changesValidators.keySet().stream().
-                filter(trigger -> trigger.triggeredByField(entityField)).
-                flatMap(trigger -> changesValidators.get(trigger).stream());
+    private Stream<EntityChangeValidator<E>> findValidatorsTriggeredByFields(Collection<? extends EntityField<E, ?>> entityFields, ChangeOperation changeOperation) {
+        return changesValidators.stream().
+                filter(validator -> validator.v1().triggeredByFields(entityFields)).
```
effiban, line 139: Please rename `validator` to `validatorPair` or `validatorAndTrigger`  - to make it less confusing

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidator.java
```diff
@@ -12,17 +10,20 @@
...
 import java.util.*;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static java.util.stream.Collectors.toSet;
 
 public class EntityChangeCompositeValidator<E extends EntityType<E>> implements ChangesValidator<E> {
 
-    private final ValidationTrigger<E> ANY_FIELD_TRIGGER = entityField -> true;
+    private final ValidationTrigger<E> ANY_CHANGE_TRIGGER = entityField -> true;
```
effiban, line 24: Let's rename the lambda param to `entityFields` for clarity (it's a little easier to understand that way, because you refactored the interface)

effiban (approved):

main/src/test/java/com/kenshoo/pl/entity/internal/validators/FieldTriggerTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import static org.junit.Assert.*;
+
+@RunWith(MockitoJUnitRunner.Silent.class)
+public class FieldTriggerTest {
+
+    private final FieldTrigger fieldTrigger = new FieldTrigger<>(TestEntity.FIELD_1);
+
+    @Test
+    public void triggered_exactly_by_field() {
```
galkoren, line 18: Hasson convinced me to use cammelCase on test methods

main/src/test/java/com/kenshoo/pl/entity/spi/helpers/EntityChangeCompositeValidatorTest.java
```diff
@@ -184,4 +184,21 @@ public void registerAncestorsValidatorForUpdateTest() {
         validator.validate(entityChanges, ChangeOperation.UPDATE, changeContext);
         verify(ancestorsValidator).validate(any(CurrentEntityState.class));
     }
+
+    @Test
+    public void registerAncestorsValidatorForEmptyUpdateTest() {
```
galkoren, line 189: Let the test name clearly describe the requirement by stating what is expected and when.

> ancestor validator should be invoked even when nothing is changed

main/src/test/java/com/kenshoo/pl/entity/internal/validators/FieldTriggerTest.java
effiban: In JUnit 5 we can add descriptions for tests and then the method name will not be so important. I think we can upgrade, will check soon.
In any case, even though it is not the Java best practice underscores are valid and I think the readability here is more important.
See here for example:
https://softwareengineering.stackexchange.com/questions/237561/naming-test-methods-in-java

dimagorelik:
@galkoren 
I fixed the test according to the review. It is hard to implement toggle and I think we can release the fix without it. Please, approve my change.

PR #185, opened 2020-08-24
PL-99: Add Triptional methods for examining the value

effiban:
To support a use-case that has already come up in Kenshoo search - adding three methods for checking the value of a `Triptional`.
`filter()` already exists in `Optional`.
The other two exist in the Scala `Option`. We chose different names:
- `equalsValue()` instead of `contains()`
- `matches()` instead of `exists()`

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/Triptional.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/Triptional.java
@@ -5,10 +5,7 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.function.BiFunction;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Supplier;
+import java.util.function.*;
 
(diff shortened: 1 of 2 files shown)


effiban:
@amirz

main/src/main/java/com/kenshoo/pl/entity/Triptional.java
```diff
@@ -314,6 +311,50 @@ public void ifNotNull(final Consumer<? super T> consumer) {
...
+    /**
+     * Return {@code true} if there is a value present (possibly {@code null), and it equals the input;
+     * otherwise {@code false}
+     *
+     * @param value a value to compare to this value
+     * @return {@code true} if there is a value present, and it equals the input; otherwise {@code false}
+     */
+    public boolean contains(final T value) {
+        return isPresent() && Objects.equals(this.value, value);
+    }
```
effiban, line 323: For completeness - there should also be an option to specify an equality function like `equals()` has.
Leaving that for a PR later on

effiban:
@galkoren - comments fixed (rename `contains()` to `equalsValue()` and add test for both `null`)

galkoren (approved):

PR #186, opened 2020-08-30
Migration simulator 

galkoren:
This service currently exists in the KS and only for campaigns.
I want to write it again in a generic way and to be covered by tests.

So what is it?
For migration purpose, we want to run the old service (`ActualDatabaseMutator`) side-by-side with Persistence Layer without output generators.
The output of the simulator is a diff between the actual run and the simulated run.
It goes like this:
1. Fetch initial state (Only for UPDATE commands. Not needed for CREATE).
1. Run PL (without output gen) and the old flow.
1. Fetch final state.
1. Compare the values in PL commands (after enrichers were triggered) with the diff of FinalState - InitialState.

* This PR is only for the simpler **creation** flow. The update flow shall come next.


main/src/main/java/com/kenshoo/pl/migration/RealMutatorError.java
```diff
@@ -0,0 +1,22 @@
+package com.kenshoo.pl.migration;
+
+import com.kenshoo.pl.entity.EntityType;
+import com.kenshoo.pl.entity.Identifier;
+
+public class RealMutatorError<E extends EntityType<E>, ID extends Identifier<E>> {
```
tzachz, line 6: perhaps rename to `ActualMutatorError` just to align with terminology used in `ActualDatabaseMutator`

main/src/main/java/com/kenshoo/pl/migration/DualRunSimulator.java
```diff
@@ -0,0 +1,131 @@
...
+        if (simulatedResult.errors.isEmpty() && realMutationError.isPresent()) {
+            return Optional.of(new ComparisonMismatch<>(simulatedResult.id, "Simulated mutation was successful but real mutation finished with the following error: " + realMutationError.get().getDescription()));
+        }
+
+        if (simulatedResult.errors.size() > 0 && realMutationError.isEmpty()) {
+            return Optional.of(new ComparisonMismatch<>(simulatedResult.id, "Real mutation was successful but simulated mutation finished with the following errors: " + simulatedResult.errors));
+        }
+
+        if (databaseFinalState == null) {
+            return Optional.of(new ComparisonMismatch<>(simulatedResult.id, "Could not found real mutation for the simulated command"));
```
tzachz, line 76:
```suggestion
            return Optional.of(new ComparisonMismatch<>(simulatedResult.id, "Could not find real mutation for the simulated command"));
```

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
```diff
@@ -280,7 +280,8 @@ private boolean isMissing(ChangeEntityCommand<?> cmd, ChangeContext context) {
     private void populateIdentityField(final ChangeEntityCommand<ROOT> cmd, final ChangeContext changeContext, final EntityField<ROOT, Object> idField) {
         final CurrentEntityState currentState = Optional.ofNullable(changeContext.getEntity(cmd))
                                       .orElseThrow(() -> new IllegalStateException("Could not find entity of command in the change context"));
-        cmd.set(idField,  currentState.get(idField));
+        Object value = cmd.containsField(idField) ? cmd.get(idField) : currentState.get(idField);
```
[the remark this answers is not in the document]
galkoren: Before this fix, when command included the ID as a field change, then PL crashed with exception.
This is why it won't cause a regression.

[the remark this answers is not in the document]
galkoren: But thinking again, maybe the fix should not be in here...
I will extract it to a new PR.

tzachz (approved):
LGTM :+1:
(would appreciate if other reviewers who are more familiar with this code can take a look)

effiban (commented):
Good job overall, the logic is clear and properly divided into small classes.
I have some style comments.

The integration tests have good coverage, however I still think (as I always do) that there should be unit tests for every class - including `ActualResultFetcher` and `InitialStateRecorder`, but I'll leave it up to you.

main/src/main/java/com/kenshoo/pl/entity/PersistenceLayer.java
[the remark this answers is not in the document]
dimagorelik: @galkoren 
I agree it is a bit risky.
I am not sure that I understand the flow 100%.

main/src/main/java/com/kenshoo/pl/simulation/internal/InitialStateRecorder.java
```diff
@@ -0,0 +1,39 @@
...
+        this.entities = commands.stream().collect(toMap(EntityChange::getIdentifier, ctx::getEntity));
+    }
+
+    @Override
+    public Stream<? extends EntityField<?, ?>> requiredFields(Collection<? extends EntityField<E, ?>> fieldsToUpdate, ChangeOperation changeOperation) {
+        return requiredFields();
+    }
+
+    public Entity get(Identifier<E> id) {
+        return entities.get(id);
```
dimagorelik, line 32: @galkoren 
There are some scenarios when the entity would not exist for the requested id

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
```diff
@@ -0,0 +1,139 @@
...
+ * modification with the content of the simulated commands.
+ *
+ * <b>Limitations</b>
+ * + One-to-many relations: Child commands are not supported.
+ * + ID must be provided to all commands, including CREATE commands, to enable matching them with the actual DB mutation.
+ *
+ * @param <E> EntityType
+ * @param <ID> ID type
+ */
+public class DualRunSimulator<E extends EntityType<E>, ID extends Identifier<E>> {
```
dimagorelik, line 36: @galkoren 
Currently, It does not support one to many scenarios, right?

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
```diff
@@ -0,0 +1,139 @@
...
+     * Note that auto-increment is not supported for simulation. IDs must be provided in order to compare simulated
+     * commands with actual mutations.
+     *
+     * @param databaseMutator
+     * @param commandsToSimulate
+     * @return
+     */
+    public List<ComparisonMismatch<E, ID>> runCreation(
+            ActualDatabaseMutator<E, ID> databaseMutator,
+            Collection<? extends Pair<ID, ? extends CreateEntityCommand<E>>> commandsToSimulate) {
```
dimagorelik, line 62: @galkoren 
Why should we receive List<Pair<ID, Command>> instead of List<CreateEntityCommand<E>>?

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
```diff
@@ -0,0 +1,139 @@
...
+
+        final var plFlow = flowToSimulate.withoutOutputGenerators().build();
+
+        populateIdsInCreateCommands(commandsToSimulate);
```
dimagorelik, line 66: @galkoren 
I do not understand this population

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
dimagorelik: @galkoren 
Maybe we should define this API for creating simulation

    public List<ComparisonMismatch<E, ID>> runCreation(
            UniqueKey<ID> comparisonKey,
            ActualDatabaseMutator<E, ID> databaseMutator,
            Collection<? extends CreateEntityCommand<E>> commandsToSimulate)

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
```diff
@@ -0,0 +1,139 @@
...
+
```
dimagorelik, line 67: @galkoren 
Catch  all exceptions and continue flow

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
```diff
@@ -0,0 +1,139 @@
...
+        final var simulatedResults = seq(pl.create(values(commandsToSimulate), plFlow).getChangeResults())
+                .zip(commandsToSimulate)
+                .map(r -> new SimulatedResult<>(r.v1.getCommand(), r.v2.getKey(), r.v1.getErrors()))
+                .collect(toList());
+
+        final var actualErrors = seq(databaseMutator.run()).toMap(__ -> __.getId());
+
```
dimagorelik, line 74: @galkoren 
Catch all exceptions, write warnings to log, and return empty result.

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Note that auto-increment is not supported for simulation. IDs must be provided in order to compare simulated
+     * commands with actual mutations.
+     *
+     * @param databaseMutator
+     * @param commandsToSimulate
+     * @return
+     */
+    public List<ComparisonMismatch<E, ID>> runCreation(
```
dimagorelik, line 60: @galkoren 
I think it will be helpful also to measure the simulation time agaist mutation time. It could be an indication about significant performance degradetion.

effiban: @dimagorelik - how can we do this if the simulation doesn't write to the DB?

dimagorelik: @effiban 
It is not for comparisan between the new and old flow. It is only for performance degradation. For example, we should stop simulation if it causes to flow 30% slowness.

effiban: I don't understand - how are you measuring the time? If you measure the old flow including the DB write, then simulation will always be faster...

main/src/main/java/com/kenshoo/pl/simulation/DualRunSimulator.java
```diff
@@ -50,24 +46,28 @@ public DualRunSimulator(PLContext plContext, ChangeFlowConfig.Builder<E> flowToS
...
      * @param databaseMutator
-     * @param commandsToSimulate
+     * @param commandsToSimulate - must contain values for the unique key.
      * @return
      */
-    public List<ComparisonMismatch<E, ID>> runCreation(
-            ActualDatabaseMutator<E, ID> databaseMutator,
-            Collection<? extends Pair<ID, ? extends CreateEntityCommand<E>>> commandsToSimulate) {
+    public List<ComparisonMismatch<E>> runCreation(
+            UniqueKey<E> uniqueKey,
```
galkoren, line 60: @dimagorelik as suggested

dimagorelik:
@galkoren 
Now, It looks good

dimagorelik:
@galkoren 
Please, add exception handlinging

dimagorelik (approved):

main/src/main/java/com/kenshoo/pl/simulation/internal/ResultComparator.java
```diff
@@ -0,0 +1,74 @@
...
+
+        final var mismatchingFields = inspectedFields.stream()
+                .map(field -> getFieldMismatch(field, simulatedResult.getCommand(), actualResult))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(toList());
+
+        return mismatchingFields.isEmpty()
+                ? Optional.empty()
+                : Optional.of(new ComparisonMismatch<>(simulatedResult.getId(), "Found field mismatch: " + mismatchingFields));
```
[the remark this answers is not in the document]
galkoren: I tried. The compiler type inference wasn't strong enough to handle it.

PR #188, opened 2020-09-08
Fix required field adapter

dimagorelik:
Fix bug in the required field adapter.
The validation should be triggered always for creation command.

### DIFF
--- a/main/src/test/java/com/kenshoo/pl/entity/internal/validators/RequiredFieldValidationAdapterTest.java
+++ b/main/src/test/java/com/kenshoo/pl/entity/internal/validators/RequiredFieldValidationAdapterTest.java
@@ -85,6 +85,11 @@ public class RequiredFieldValidationAdapterTest {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


main/src/test/java/com/kenshoo/pl/entity/internal/validators/RequiredFieldValidationAdapterTest.java
```diff
@@ -84,6 +84,11 @@ public void triggered_by_required_field() {
         assertTrue(underTest.trigger().triggeredByFields(List.of(TestEntity.FIELD_1)));
     }
 
+    @Test
+    public void triggered_by_any_required_field() {
```
effiban, line 88: I would just call it "triggered_by_any_field" - the word "required" is confusing

dimagorelik: @effiban 
Fixed

effiban (approved):

PR #189, opened 2020-09-09
fix shared children state on CREATE

galkoren:
This PR
---
Children of CREATE operator share the same CurrentState.
When different children of different parents require fields from their parents, it doesn't work.
EntitiesToContextFetcher populates the parent values into the children, but since they all share the same state, the values override each other.

Test that fails before the fix:
```java
    @Test
    public void create_child_with_supplier_from_parent() {

        insert(newParent().with(NAME, "moshe"));
        insert(newParent().with(NAME, "david"));

        var supplier = fromOldValue(ParentEntity.NAME, parentName -> "I'm the child of " + parentName);

        update(
            existingParentWithId(generatedId(0)).with(insertChild().with(ORDINAL, 0).with(FIELD_1, supplier)),
            existingParentWithId(generatedId(1)).with(insertChild().with(ORDINAL, 0).with(FIELD_1, supplier))
        );
```

Expecting "child of moshe" and "child of david", but actual results are "child of david" and "child of david".

The fix is simple: each child command shall have it's own state.
To reduce memory footprint, I also optimized the `map` inside the CurrentEntityMutableState to be null until it is really populated.

Next optional step
---
Memory footprint could be farther reduced because, as for today, each child has it's own copy of the ancestor values while they can actually share it.


main/src/main/java/com/kenshoo/pl/entity/CurrentEntityMutableState.java
```diff
@@ -9,15 +9,18 @@
 
 public class CurrentEntityMutableState implements CurrentEntityState {
 
-    private final Map<EntityField<?, ?>, Object> fields = new HashMap<>();
-    private Map<EntityType, List<FieldsValueMap>> manyByType;
+    private Map<EntityField<?, ?>, Object> fields = null;
```
galkoren, line 12: being lazy in order to reduce memory footprint

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesToContextFetcher.java
```diff
@@ -106,8 +105,7 @@ private Object getValue(ChangeContext context, ChangeEntityCommand cmd, EntityFi
         E entityType = flowConfig.getEntityType();
         Collection<EntityField<E, ?>> foreignKeys = entityType.determineForeignKeys(flowConfig.getRequiredRelationFields()).filter(not(new IsFieldReferringToParent<>(context.getHierarchy(), entityType))).collect(toList());
         if (foreignKeys.isEmpty()) {
-            CurrentEntityState sharedEntity = new CurrentEntityMutableState();
```
galkoren, line 109: no more shared state

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesToContextFetcher.java
```diff
@@ -106,8 +104,10 @@ private Object getValue(ChangeContext context, ChangeEntityCommand cmd, EntityFi
...
-            commands.forEach(cmd -> context.addEntity(cmd, sharedEntity));
+            seq(commands).grouped(firstAncestorHoldingRequiredFieldsFor(entityType, context)).forEach(group -> {
```
galkoren, line 107: please review again

main/src/main/java/com/kenshoo/pl/entity/Hierarchy.java
```diff
@@ -48,6 +48,10 @@ boolean contains(EntityField<?, ?> field) {
         );
     }
 
+    public int distanceBetween(EntityType<?> child, EntityType<?> parent) {
+        return parent == child ? 0 : 1 + distanceBetween(getParent(child).get(), parent);
```
effiban, line 52: 1) This is not in functional style (and unsafe) because you call `Optional::get` (you assume it can't fail).
2) If called by accident with a child which is not a descendant of the parent, it will throw an exception with no message

I would do it like this:
```
public Optional<Integer> distanceBetween(EntityType<?> child, EntityType<?> ancestor) {
    if (child == ancestor) {
        return Optional.of(0);
    }
    return getParent(child)
             .flatMap(parent -> distanceBetween(parent, ancestor))
             .map(dist -> dist+1);
}
```

The calling code eventually uses an `Optional` so this should fit well

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesToContextFetcher.java
```diff
@@ -116,6 +116,22 @@ private Object getValue(ChangeContext context, ChangeEntityCommand cmd, EntityFi
...
+                .mapToInt(requiredField -> ctx.getHierarchy().distanceBetween(currentLevel, requiredField.getWhereToQuery()))
+                .min();
+
+        return distance.isPresent()
+                ? child -> Optional.of(getAncestor(child, distance.getAsInt()))
+                : child -> Optional.empty();
+    }
+
+    private ChangeEntityCommand<?> getAncestor(ChangeEntityCommand<?> cmd, int distance) {
+        return distance == 1 ? cmd.getParent() : getAncestor(cmd.getParent(), distance - 1);
```
effiban, line 132: Here too you can make it functional and safe,  since you need an `Optional` anyway:

```
private Optional<ChangeEntityCommand<?>> getAncestor(ChangeEntityCommand<?> cmd, int distance) {
    final var optionalParent = Optional.ofNullable(cmd.getParent());
    if (distance == 1) {
        return optionalParent;
    }
    return optionalParent.flatMap(parent -> getAncestor(parent, distance - 1));
}
```

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesToContextFetcher.java
```diff
@@ -106,8 +104,10 @@ private Object getValue(ChangeContext context, ChangeEntityCommand cmd, EntityFi
         E entityType = flowConfig.getEntityType();
         Collection<EntityField<E, ?>> foreignKeys = entityType.determineForeignKeys(flowConfig.getRequiredRelationFields()).filter(not(new IsFieldReferringToParent<>(context.getHierarchy(), entityType))).collect(toList());
         if (foreignKeys.isEmpty()) {
-            CurrentEntityState sharedEntity = new CurrentEntityMutableState();
-            commands.forEach(cmd -> context.addEntity(cmd, sharedEntity));
+            seq(commands).grouped(firstAncestorHoldingRequiredFieldsFor(entityType, context)).forEach(group -> {
+                CurrentEntityState sharedEntity = new CurrentEntityMutableState();
+                group.v2.forEach(cmd -> context.addEntity(cmd, sharedEntity));
```
galkoren, line 109: instead of this, let populate to enrich

dimagorelik:
@galkoren 
It is a very complicated change. Let's do the review in zoom. Anyway, we need toggle here.

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesToContextFetcher.java
```diff
@@ -34,70 +39,23 @@ public EntitiesToContextFetcher(EntitiesFetcher entitiesFetcher) {
...
+        populateFieldsFromAllParents(seq(commands).filter(notMissing(context)).toList(), context);
     }
 
-    private Predicate<FieldFetchRequest> queriedOn(EntityType e) {
-        return r -> r.getWhereToQuery().equals(e);
+    private <E extends EntityType<E>> void populateFieldsFromAllParents(List<? extends ChangeEntityCommand<E>> commands, ChangeContext ctx) {
+        seq(commands)
+                .zip(ancestorStateCollector.collect(commands, ctx))
+                .filter(pair -> pair.v2.isPresent())
+                .forEach(consumer((cmd, ancestorState) ->  ctx.addEntity(cmd, new OverridingEntity(ancestorState.get(), ctx.getEntity(cmd)))));
```
galkoren, line 52: don't need another override

main/src/main/java/com/kenshoo/pl/entity/internal/EntitiesToContextFetcher.java
```diff
@@ -10,13 +11,17 @@
...
+import static java.util.function.Predicate.not;
 import static java.util.stream.Collectors.*;
 import static org.jooq.lambda.Seq.seq;
+import static org.jooq.lambda.tuple.Tuple.consumer;
 
 
 public class EntitiesToContextFetcher {
 
     private final EntitiesFetcher entitiesFetcher;
+    private final AncestorStateCollector ancestorStateCollector = AncestorStateCollector.INSTANCE;
```
galkoren, line 23: replace new `new`

effiban: what does this mean?

galkoren:
This was proved to be slower than the simple fix.

PR #191, opened 2020-09-13
Improve changes validator api

dimagorelik:
Improve changes validator API


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/OptimizedChangesValidator.java
```diff
@@ -0,0 +1,166 @@
...
+import com.kenshoo.pl.entity.EntityType;
+import com.kenshoo.pl.entity.internal.EntityTypeReflectionUtil;
+import com.kenshoo.pl.entity.internal.validators.*;
+import com.kenshoo.pl.entity.spi.*;
+
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class OptimizedChangesValidator<E extends EntityType<E>> implements ChangesValidator<E> {
```
dimagorelik, line 18: @galkoren @effiban 
Do you like this name? Do you have another suggestion?

effiban: I don't understand why you call it "optimized" - what is the optimization ?

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/OptimizedChangesValidator.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    public Stream<EntityField<?, ?>> requiredFields(Collection<? extends EntityField<E, ?>> fieldsToUpdate, ChangeOperation changeOperation) {
+        return findValidatorsTriggeredByFields(fieldsToUpdate, changeOperation)
+                .flatMap(ChangeValidatorAdapter::fieldsToFetch);
+    }
+
+    public void register(ChangeValidatorAdapter<E> validatorAdapter) {
+        triggeredChangeValidators.add(validatorAdapter);
+    }
```
effiban, line 46: Shouldn't this be inside the builder as well?

dimagorelik: @effiban 
Fixed

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/OptimizedChangesValidator.java
galkoren: Name is tricky... this class is yet another adaptor, an adaptor of all other adaptors to become a single Validator.
So it is a composite. But we already have many other composites. We need to discuss it.

effiban: As I understand the difference from the previous impl is that we now have a fluent way to add validators - right?
If so we can call it FluentCompositeChangeValidator

dimagorelik: @effiban 
Please, pay attention that this validator implements the general ChangesValidator interface

The validator does the number of optimization:
1. Calculate fields to fetch according to updated fields
2. Apply only validators that triggered by updated fields

effiban: Ok, but you didn't actually change the logic compare to the old name, right? You just added the builder?

effiban: I don't think that it is more or less "optimized" than anything else - all the validators should calculate fields to fetch and be applied only when needed.
It is composite like Gal said, and it also uses adapters.
So maybe we could call it `CompositeAdaptedChangeValidator` or something like that?

dimagorelik: @effiban 
I didn't do any change

dimagorelik: @effiban 
This name CompositeAdaptedChangeValidator does not say to the user why who need it

effiban: @dimagorelik - so I think `CompositeChangeValidator` should be enough, and there should eventually only be one composite (I'm not sure but I think we can get rid of "compound" eventually

dimagorelik: @effiban 
I can explain why we need 2 such validators. Let's speak

dimagorelik: @effiban @galkoren 
One more name suggestion - TriggeredByChangesValidator

PR #193, opened 2020-09-15
dont crash when ID not generated because duplicate ignored

galkoren:
Currently, when using `auto-inc` + `CreateRecordCommand.IGNORE`, and the change is really ignored, then PL is crashed with an exception when trying to populate the ID back to the Identifier and then to the change.


main/src/main/java/com/kenshoo/pl/entity/internal/DbCommandsOutputGenerator.java
```diff
@@ -83,9 +83,13 @@ private void populateGeneratedIdsToContext(final EntityField<E, Object> identity
                 .map(change -> ImmutablePair.of(change, changesContainer.getInsert(entityType.getPrimaryTable(), change)))
                 .filter(pair -> pair.getRight().isPresent())
                 .forEach(pair -> {
-                    final CreateRecordCommand cmd = pair.getRight().get();
-                    Object generatedValue = cmd.get(identityTableField);
-                    changeContext.addEntity(pair.getLeft(), new EntityWithGeneratedId(identityField, generatedValue));
+                    if (pair.getRight().get().get(identityTableField) == null) {
+                        changeContext.addValidationError(pair.left, new ValidationError("Auto generated key was not retrieved"));
```
dimagorelik, line 87: @galkoren 
I am not sure that it is an error. Maybe we should simply ignore it?

dimagorelik: @galkoren 
What happens if the ignored command has a child command?

galkoren: This PR doesn't fix child commands. My guess is that the populator will crash.
I'm only doing this fix because it is required by ConvUp.
It is a good question whether it should be an error or not... Maybe an error is good because it gives indication to the user and later we can use it to stop generating children (so that the fix will be recursive).

PR #194, opened 2020-09-22
Delete compound changes validator factory

dimagorelik:
Delete compound changes validator factory


dimagorelik:
@galkoren 
Should fix our wikis and books?

effiban (approved):

PR #196, opened 2020-09-23
Register required field validator according to annotation

dimagorelik:
The pospouse of the change is to unify the validation mechanism of the required fields.

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
@@ -8,6 +8,7 @@ import com.kenshoo.pl.entity.internal.audit.*;
 import com.kenshoo.pl.entity.spi.*;
 import com.kenshoo.pl.entity.spi.helpers.EntityChangeCompositeValidator;
 import com.kenshoo.pl.entity.spi.helpers.ImmutableFieldValidatorImpl;
+import com.kenshoo.pl.entity.spi.helpers.RequiredFieldValidatorImpl;
 import org.jooq.lambda.Seq;
 
 import java.util.*;
@@ -58,7 +59,7 @@ public class ChangeFlowConfig<E extends EntityType<E>> {
         this.requiredFields = requiredFields;
         this.childFlows = childFlows;
         this.postFetchFilters = ImmutableList.of(new MissingParentEntitiesFilter<>(entityType.determineForeignKeys(requiredRelationFields).collect(toList())), new MissingEntitiesFilter<>(entityType));
-        this.postSupplyFilters = ImmutableList.of(new RequiredFieldsChangesFilter<>(requiredFields));
+        this.postSupplyFilters = features.isEnabled(Feature.RequiredFieldValidator) ? Collections.emptyList() : ImmutableList.of(new RequiredFieldsChangesFilter<>(requiredFields));
         this.retryer = retryer;
         this.auditRequiredFieldsCalculator = auditRequiredFieldsCalculator;
         this.auditRecordGenerator = auditRecordGenerator;
@@ -257,8 +258,14 @@ public class ChangeFlowConfig<E extends EntityType<E>> {
             requiredRelationFields.collect(toCollection(() -> this.requiredRelationFields));
         }
 
-        /* not public */ void withRequiredFields(Stream<EntityField<E, ?>> requiredFields) {
+        /* not public */ void withDeprecatedRequiredFields(Stream<EntityField<E, ?>> requiredFields) {
             requiredFields.collect(toCollection(() -> this.requiredFields));
+        }
+
+        /* not public */ void withRequiredFields(Stream<EntityField<E, ?>> requiredFields) {
+            EntityChangeCompositeValidator<E> compositeValidator = new EntityChangeCompositeValidator<>();
+            requiredFields.forEach(requiredField -> compositeValidator.register(new RequiredFieldValidatorImpl<>(requiredField, Errors.FIELD_IS_REQUIRED)));
+            this.withValidator(compositeValidator);
         }
 
         /* not public */ void withImmutableFields(Stream<EntityField<E, ?>> immutableFields) {
--- a/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
@@ -8,7 +8,7 @@ import com.kenshoo.pl.entity.internal.audit.*;
 import com.kenshoo.pl.entity.spi.*;
 import com.kenshoo.pl.entity.spi.helpers.EntityChangeCompositeValidator;
 import com.kenshoo.pl.entity.spi.helpers.ImmutableFieldValidatorImpl;
-import com.kenshoo.pl.entity.spi.helpers.RequiredFieldValidatorImpl;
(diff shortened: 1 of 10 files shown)


main/src/main/java/com/kenshoo/pl/entity/spi/helpers/RequiredFieldValidatorImpl.java
```diff
@@ -0,0 +1,27 @@
+package com.kenshoo.pl.entity.spi.helpers;
+
+import com.kenshoo.pl.entity.EntityField;
+import com.kenshoo.pl.entity.EntityType;
+import com.kenshoo.pl.entity.spi.RequiredFieldValidator;
+
+public class RequiredFieldValidatorImpl<E extends EntityType<E>, T> implements RequiredFieldValidator<E, T> {
```
effiban, line 7: This is a simple implementation but the clients (KS) have more complicated ones. Assuming that we want to keep allowing such complicated impls, and this one is specifically for the annotation - then maybe give it a name like `SimpleRequiredFieldValidator`, `OnlyRequiredFieldValidator`

main/src/test/java/com/kenshoo/pl/entity/ChangeFlowConfigBuilderFactoryTest.java
```diff
@@ -17,14 +18,21 @@
...
 
     @Test
     public void testFieldsAnnotatedWithDontPurgeArePassedToTheFalseUpdatePurger() {
         assertThat(getPurger(buildFlow()).getFieldsToRetain(), containsInAnyOrder(MockedEntity.INSTANCE.field1));
     }
 
     private ChangeFlowConfig<MockedEntity> buildFlow() {
-        return ChangeFlowConfigBuilderFactory.newInstance(mock(PLContext.class), MockedEntity.INSTANCE).build();
+        PLContext plContext = mock(PLContext.class);
+        when(plContext.generateFeatureSet()).thenReturn(FeatureSet.EMPTY);
```
effiban, line 34: Please add tests for the case when the feature is enabled

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -58,7 +59,7 @@ private ChangeFlowConfig(E entityType,
         this.requiredFields = requiredFields;
         this.childFlows = childFlows;
         this.postFetchFilters = ImmutableList.of(new MissingParentEntitiesFilter<>(entityType.determineForeignKeys(requiredRelationFields).collect(toList())), new MissingEntitiesFilter<>(entityType));
-        this.postSupplyFilters = ImmutableList.of(new RequiredFieldsChangesFilter<>(requiredFields));
+        this.postSupplyFilters = features.isEnabled(Feature.RequiredFieldValidator) ? Collections.emptyList() : ImmutableList.of(new RequiredFieldsChangesFilter<>(requiredFields));
```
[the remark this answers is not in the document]
dimagorelik: @effiban 
The  RequiredFieldsChangeFilter is not fully depreciated. It is used also in another place

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/RequiredFieldValidatorImpl.java
dimagorelik: @effiban 
I am not sure that we should call it Simple or Only.

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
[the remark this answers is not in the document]
effiban: @dimagorelik  - ok, I see it

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/RequiredFieldValidatorImpl.java
effiban: Usually we use `Impl` for the class name when there is only one impl (or only one "real" non-mock impl). 
It becomes less clear when you allow multiple impls that are all production code  - so I prefer a different and more specific name.

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -257,10 +258,16 @@ public Builder(E entityType) {
             requiredRelationFields.collect(toCollection(() -> this.requiredRelationFields));
         }
 
-        /* not public */ void withRequiredFields(Stream<EntityField<E, ?>> requiredFields) {
+        /* not public */ void withDeprecatedRequiredFields(Stream<EntityField<E, ?>> requiredFields) {
```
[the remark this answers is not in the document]
galkoren: It's not public :)
Why not deleting it completely?

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
```diff
@@ -257,10 +258,16 @@ public Builder(E entityType) {
...
             requiredFields.collect(toCollection(() -> this.requiredFields));
         }
 
+        /* not public */ void withRequiredFields(Stream<EntityField<E, ?>> requiredFields) {
+            EntityChangeCompositeValidator<E> compositeValidator = new EntityChangeCompositeValidator<>();
```
galkoren, line 266: Let's add a utility static method (`EntityChangeCompositeValidator.compositeOf()`) to composite this.
Usage could be nice like this:
```java
/* not public */ void withRequiredFields(Stream<EntityField<E, ?>> requiredFields) {
    return compositeOf(requiredFields.map(RequiredFieldValidatorImpl::new));
}
```

main/src/main/java/com/kenshoo/pl/entity/spi/helpers/SimpleRequiredFieldValidator.java
```diff
@@ -0,0 +1,27 @@
+package com.kenshoo.pl.entity.spi.helpers;
+
+import com.kenshoo.pl.entity.EntityField;
+import com.kenshoo.pl.entity.EntityType;
+import com.kenshoo.pl.entity.spi.RequiredFieldValidator;
+
+public class SimpleRequiredFieldValidator<E extends EntityType<E>, T> implements RequiredFieldValidator<E, T> {
```
galkoren, line 7: I remember we didn't want to expose such validators to the user because we want them to implement an interface rather than we give them builders.
Also, this is not an `spi` (it is not an interface).
Let's make it a package private class close to the flow builder.

dimagorelik: @galkoren 
I think this class could be used in the client code.  It is a default implementation.

main/src/main/java/com/kenshoo/pl/entity/ChangeFlowConfig.java
dimagorelik: @galkoren 
Is the builder preferable?
I created it in the different PR.

dimagorelik: @galkoren 
I prefer to do EntityChangeCompositeValidator.compositeOf()  in differnt PR

effiban (approved):

PR #197, opened 2020-09-30
Unique key value refactoring

dimagorelik:
Fix Unique Key Value interface


effiban (approved):
Looks good. I have style comments - but they are related to code that you moved, so can be done in separate PR

PR #200, opened 2020-10-05
Define key interface

dimagorelik:
Define key interface


main/src/main/java/com/kenshoo/pl/entity/IdentifierDefinition.java
```diff
@@ -0,0 +1,51 @@
...
+
+    static <E extends EntityType<E>, T1, T2, T3, T4, T5> QuintupleUniqueKey<E, T1, T2, T3, T4, T5> uniqueKey(
+            EntityField<E, T1> f1,
+            EntityField<E, T2> f2,
+            EntityField<E, T3> f3,
+            EntityField<E, T4> f4,
+            EntityField<E, T5> f5) {
+        return new QuintupleUniqueKey<>(f1, f2, f3, f4, f5);
+    }
+}
```
effiban, line 51: This looks good, however I don't think it's good practice to put static methods (which are factory methods) in an interface. Instead I would extract all of the static methods to a new class called `Identifiers` or `UniqueKeys`

main/src/main/java/com/kenshoo/pl/entity/IdentifierDefinition.java
```diff
@@ -0,0 +1,51 @@
+package com.kenshoo.pl.entity;
+
+import org.jooq.Record;
+import org.jooq.TableField;
+
+import java.util.List;
+
+public interface IdentifierDefinition<E extends EntityType<E>> {
```
effiban, line 8: I would call it `IdentifierType`

main/src/main/java/com/kenshoo/pl/entity/IdentifierDefinition.java
dimagorelik: @effiban 
I took an example from Tupple implementation. I think it is a known pattern.

main/src/main/java/com/kenshoo/pl/entity/IdentifierDefinition.java
dimagorelik: @effiban 
I fix it

main/src/main/java/com/kenshoo/pl/entity/IdentifierDefinition.java
effiban: @dimagorelik - you're right, it is used in `Tuple` and is an acceptable pattern - I personally don't like it so much because you are mixing together 2 type of methods in the same file - the abstract interface methods and the static ones (I like to keep each type focused and as small as possible).
See the discussion here: https://www.baeldung.com/java-static-default-methods. 
At the very bottom (Conclusion) he mentions that it doesn't look so good from the object-oriented perspective and is mostly a solution for keeping backwards compatibility (I guess he means in case you don't want to add any new classes to your API).

Anyway - not critical for me either way.

dimagorelik:
@effiban 
Fixed

main/src/main/java/com/kenshoo/pl/entity/IdentifierDefinition.java
dimagorelik: @effiban 
Ok, I will read

main/src/main/java/com/kenshoo/pl/entity/IdentifierDefinition.java
dimagorelik: @effiban 
fixed

effiban (approved):

PR #202, opened 2020-10-12
Unique key refactor

dimagorelik:
New API is added to IdentifyerType createIdentifier(Entity entity)
UniqueKeyValue became protected class

### DIFF
--- a/main/src/main/java/com/kenshoo/pl/entity/UniqueKey.java
+++ b/main/src/main/java/com/kenshoo/pl/entity/UniqueKey.java
@@ -3,6 +3,7 @@ package com.kenshoo.pl.entity;
 import com.google.common.collect.ImmutableList;
 import org.jooq.Record;
 import org.jooq.TableField;
+import org.jooq.lambda.Seq;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


main/src/main/java/com/kenshoo/pl/entity/UniqueKey.java
```diff
@@ -54,6 +54,15 @@ public UniqueKey(Iterable<? extends EntityField<E, ?>> fields) {
...
     }
 
+    @Override
+    public Identifier<E> createIdentifier(Entity entity) {
+        Object[] values = new Object[fields.length];
+        for (int i = 0; i < fields.length; i++) {
+            values[i] = entity.get(fields[i]);
+        }
+        return new UniqueKeyValue<>(this, values);
+    }
```
effiban, line 64: Can we use functional style (`Stream`) for this loop instead? Is there a performance issue?

dimagorelik: @effiban 
Do you prefer this code that was in the original method?
        final Object[] ids = Seq.of(key.getFields())
                .map(entity::get)
                .toArray();

effiban: @dimagorelik - yes, I think that would be better

dimagorelik: @effiban 
fixed

effiban (approved):

